PR #1, opened 2016-06-23
Silence warning for resource leak

stefanoborini:
Fixes a warning due to a python request resource leak (from connectionpool
being GC before being cleaned up). This seems to be a request issue and I am
unsure we can do anything on our side. This PR simply silences the warning.

See https://github.com/kennethreitz/requests/issues/1882
for details about the issue itself.


itziakos:
:+1:

PR #2, opened 2016-06-23
Mount common volume for an application as read-only data volume

kitchoi:
Mount a common data volume on the host machine to the container based on information given by the database.
Since we don't yet have a database, two namedtuples (Application and Volume) are coded here, and are to be placed by ORM objects.

Note: For docker version >= 1.11, mounting a host directory that does not exist will no longer create the directory (I was surprised to find out that it does for docker version < 1.11).  To be sure, we should always check if the source path exists.

I would prefer refactoring the path existence check to the ContainerManager._start_container in a separate PR.

### DIFF
--- a/remoteappmanager/handlers/home_handler.py
+++ b/remoteappmanager/handlers/home_handler.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 import socket
 import os
 from datetime import timedelta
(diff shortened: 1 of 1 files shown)


itziakos:
:+1:

kitchoi:
Thanks!

kitchoi:
Note that the host directory needs to have a read-execute permission for `others` for this works with the current docker images.

itziakos:
> Note that the host directory needs to have a read-execute permission for others for this works with the current docker images.

Yes, it is assumed that the admin has setup the folders with the right permissions.

PR #10, opened 2016-06-23
Check if volume source exists in ContainerManager

kitchoi:
The checking for whether a volume source path exists should occur in the ContainerManager._start_container

Refactoring following #2 

This simplifies HomeHandler and makes it easier to test.

### DIFF
--- a/remoteappmanager/docker/container_manager.py
+++ b/remoteappmanager/docker/container_manager.py
@@ -174,7 +174,7 @@ class ContainerManager(LoggingMixin):
         volume_targets = [binding['bind']
                           for binding in filtered_volumes.values()]
 
-        # Log the paths that are being mounted
+        # Log the paths that are not being mounted
         if volumes.keys() - filtered_volumes.keys():
             self.log.error('Path(s) does not exist, not mounting:\n%s',
                            '\n'.join(volumes.keys() - filtered_volumes.keys()))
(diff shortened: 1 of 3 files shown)


remoteappmanager/docker/container_manager.py
```diff
@@ -162,8 +163,21 @@ def _start_container(self, user_name, image_name, volumes):
         # volumes = {volume_source: {'bind': volume_target,
         #                            'mode': volume_mode}
         volumes = volumes if volumes else {}
+
+        # Filter away the volume sources that do not exist,
+        # otherwise Docker would create non-existing host directory
+        # See Docker PR #21666
+        filtered_volumes = {source: volumes[source]
+                            for source in volumes
+                            if os.path.exists(source)}
```
stefanoborini: we probably need a more strict check. do an os.path.isdir and os.access with R_OK and W_OK.

kitchoi: In case of mounting a file instead of mounting a directory?

kitchoi: For the source path being a file, it will be mounted as a file.  It is a feature of docker that might be handy at times.  I don't see why we should stop that from happening.  Thought?

kitchoi:
Sorry, set the WIP for the wrong PR 😵

remoteappmanager/docker/container_manager.py
stefanoborini: Ok, if docker handles it transparently, then it's ok. Same for read/write I assume.

Issue #8, opened 2016-06-23
Need database (table) for knowing which user can access which containers 


itziakos:
The tentative database schema is provided bellow

![db](https://cloud.githubusercontent.com/assets/403930/16307515/63ed802c-3959-11e6-9d21-633ccea40a11.png)

This should allow the basic behaviours of mounting common shared folders and controlling the basic container access of users

stefanoborini:
What do you mean with volume: target?

PR #11, opened 2016-06-23
Initial implementation of database service

stefanoborini:
Introduces the basic implementation of the database schema.


remoteappmanager/application.py
```diff
@@ -181,11 +191,45 @@ def _reverse_proxy_init(self):
...
+        self.db = orm.Database(self.config.db_url)
+
+    def _user_init(self):
+        Session = self.db.create_session_factory()
+
+        session = Session()
+        try:
+            user = session.query(orm.User).filter_by(
+                name=self.config.user).one_or_none()
+        except MultipleResultsFound:
```
stefanoborini: Unsure if this is needed. The point is that if you get multiple users out of this query, something really wrong happened at the schema definition, or we are using a database that doesn't enforce unique keys.

remoteappmanager/application_config.py
```diff
@@ -163,11 +167,3 @@ def _transfer_config(self, options_object, metadata_identifier):
         for traitlet_name, traitlet in self.traits().items():
             if traitlet.metadata.get(metadata_identifier, False):
                 self.set_trait(traitlet_name, options_object[traitlet_name])
-
-
-_global_config = ApplicationConfig()
```
stefanoborini: We never use this.

remoteappmanager/db/orm.py
```diff
@@ -0,0 +1,100 @@
...
+        Parameters
+        ----------
+        url : url
+            A sqlalchemy url to connect to a specified database.
+        **kwargs : dict
+            Additional keys will be passed at create_engine.
+        """
+        super().__init__()
+
+        self.url = url
```
stefanoborini: How traitlet-happy do we want to be? I think it's worth it... opinions?

remoteappmanager/handlers/base_handler.py
```diff
@@ -8,7 +8,7 @@ class BaseHandler(web.RequestHandler, LoggingMixin):
     """Base class for the request handler."""
 
     def get_current_user(self):
-        return self.settings["user"]
+        return self.application.user.name
```
stefanoborini: This will return a User object in the future, but for now, we just hand out the string as before.

remoteappmanager/application.py
```diff
@@ -181,11 +191,43 @@ def _reverse_proxy_init(self):
...
+                self.log.error("Multiple results found when "
+                               "querying for username {}. This is supposedly "
+                               "impossible because the username should be a "
+                               "unique key by design.".format(
+                                   self.config.user))
+                raise
+
+            if user is None:
+                user = orm.User(name=self.config.user)
+                session.add(user)
```
kitchoi: (1) Do we definitely want to add any unfound user to the database?  Jupyterhub has a `whitelist` for making sure we only add the whitelisted users to the database
(2) If yes to (1), we want to do `session.commit()` whenever we do `session.add`.

remoteappmanager/application.py
```diff
@@ -181,11 +191,43 @@ def _reverse_proxy_init(self):
...
+
+            # make sure that the user always has at least one team: his own.
+            if len(user.teams) == 0:
+                team = orm.Team(name=self.config.user)
+                user.teams.append(team)
+                session.add(team)
```
kitchoi: Similar comment to the one above, if we are sure of adding unfound user to the database, `session.commit` after `session.add`

remoteappmanager/application.py
kitchoi: We probably also want to `log.debug` saying that we are adding the user to the database.

remoteappmanager/db/orm.py
```diff
@@ -0,0 +1,142 @@
...
+
+Base = declarative_base()
+
+
+class UserTeam(Base):
+    """ The user (n <-> n) team association table """
+    __tablename__ = "user_team"
+    id = Column(Integer, primary_key=True)
+    user_id = Column(Integer, ForeignKey('user.id'))
+    team_id = Column(Integer, ForeignKey('team.id'))
```
kitchoi: From the examples that I saw that implement many-to-many mappings assigned primary_key=True to both of these ids (`user_id` and `team_id` in this case).  There is no need for an additional `id` for the table either.  (e.g. https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/orm.py#L262)

stefanoborini: @kitchoi that makes me very uncomfortable. You don't want two primary keys. You eventually want a combined primary key. Unless sqlalchemy does some magic behind the scenes... but it's often not very clear how sqlalchemy maps to real sql

kitchoi: I tend to think that this makes bidirectional lookup quicker.  Same example is also found on sqlalchemy's documentation.

kitchoi: http://docs.sqlalchemy.org/en/latest/orm/join_conditions.html#self-referential-many-to-many

kitchoi: The majority of examples create the mapping using the `Table` class with the `Base.metadata`, instead of subclassing Base directly.  Haven't used sqlalchemy before, I would go with whatever the doc says or what others do.

stefanoborini: They are the "old style". Modern style is with declarative. They are equivalent, in the sense that it does a lot of work for you.

kitchoi: > Haven't used sqlalchemy before, I would go with whatever the doc says or what others do.

I also said this for the `pirmary_key=True` issue.

remoteappmanager/db/orm.py
```diff
@@ -0,0 +1,142 @@
...
+    users = relationship("User", secondary="user_team", backref="teams")
+
+
+class Application(Base):
+    """ Describes an application that should be available for startup """
+    __tablename__ = "application"
+    id = Column(Integer, primary_key=True)
+
+    #: The docker image name where the application can be found
+    image = Column(Unicode)
```
kitchoi: According to the schema in issue #8 (unless there are discussed changes I am not aware), info on common volume (volume source, target, mode) should be described in this table as well.

stefanoborini: I think that ApplicationPolicy table is redundant...

kitchoi: ?

remoteappmanager/db/orm.py
```diff
@@ -0,0 +1,142 @@
...
+    id = Column(Integer, primary_key=True)
+
+    team_id = Column(Integer, ForeignKey("team.id"))
+
+    application_id = Column(Integer, ForeignKey("application.id"))
+
+    application_policy_id = Column(Integer,
+                                   ForeignKey("application_policy.id"))
+
+    team = relationship("Team")
```
kitchoi: Does this work the same as `team = relationship(Team, primaryjoin=team_id==Team.id)`?

kitchoi: Similarly for `application` and `application_policy`

remoteappmanager/application.py
kitchoi: Please disregard the `commit` comment, I just saw the commit in the transaction context manager.  I'm bit slow.

remoteappmanager/application.py
```diff
@@ -1,11 +1,15 @@
 import os
-from tornado import web, gen
 from urllib import parse
-import tornado.ioloop
 
+from remoteappmanager.db.orm import transaction
```
kitchoi: A minor (subjective?) comment, could do a bit of ordering for the imports?
[standard libraries]
[other libraries]
[local package]

tests/db/test_orm.py
```diff
@@ -0,0 +1,116 @@
...
+                user.teams.append(team)
+
+            # make users 1 3 and 4 part of the company team
+            users[1].teams.append(company_team)
+            users[3].teams.append(company_team)
+            users[4].teams.append(company_team)
+
+            # verify back population
+            self.assertEqual(teams[0].users[0], users[0])
+            self.assertEqual(len(company_team.users), 3)
```
kitchoi: With the intended many-to-many mapping, one should be able to add users 1, 3 and 4 to the company team: `company_team.users.extend([users[1], users[3], users[4]])`
Is this tested too?

remoteappmanager/db/orm.py
kitchoi: Just checked on my local machine, if you add primary key to all of the columns (`id`, `user_id`, `team_id`), the SQL table would have a primary key `PRIMARY KEY (id, user_id, team_id)`
So there is no need for the `id` column.

remoteappmanager/application.py
stefanoborini: > Do we definitely want to add any unfound user to the databas

I have no idea. It's a matter of taste I guess? We need to extract the user anyway, and if we set the spawner to create the user, then we need to import the database and the orm objects from the spawner...

I personally think that once you spawn this process, you are already authenticated, and you either are already in the database, or if you are not, you should be. I would go with the solution that is less entangled.

@itziakos opinion?

itziakos: I think the user should be already in the database when starting the single user app. The single user app treats the database as read only.

itziakos: > then we need to import the database and the orm objects from the spawner...

I do not see why this is a problem?

stefanoborini: I don't like the idea of mixing the two things.

remoteappmanager/db/orm.py
itziakos: > I think that ApplicationPolicy table is redundant...

Why do you say that?

remoteappmanager/db/orm.py
stefanoborini: I don't know. I guess so?

tests/test_application.py
```diff
@@ -3,24 +3,29 @@
...
 
-from remoteappmanager.application_config import ApplicationConfig
 from remoteappmanager.application import Application
 from remoteappmanager.docker.container import Container
 from remoteappmanager.docker.container_manager import ContainerManager
+from tests import utils
 
 
-class TestApplication(testing.AsyncTestCase):
+class TestApplication(TempMixin, testing.AsyncTestCase):
```
kitchoi: Instead of mixin, I would prefer adding a method for removing the temporary directory and do `addCleanUp` in the setup, unless `TempMixin` is used often.

remoteappmanager/db/orm.py
itziakos: We sort it out offline, thanks @stefanoborini  and @kitchoi

tests/test_application.py
stefanoborini: Used as a mixin class for two unrelated tests classes (tornado async and unittest)

kitchoi:
LGTM Thanks!

Issue #15, opened 2016-06-24
Running containers without noVNC support still shown as Viewable/Stoppable


kitchoi:
Related to #6 and #14 
When a container's noVNC page is not reachable, we raise a TimeoutError because the server is not available.  However the container is up and running, the user still sees the View/Stop buttons.  The reverse proxy is not set up either so that clicking either button gives an HTTP error.

kitchoi:
Fixed via #21

Issue #14, opened 2016-06-24
Running containers not in cache when container is started without a public port


kitchoi:
Start any incompatible image not exposing a port (e.g. `quay.io/travisci/travis-python:latest`), the container is up and running but we do not keep it in our cache of running containers because the port is unavailable (an exception is raised).

From the user point of view, there is no View/Stop button which makes sense.
From the admin point of view, the container has to be stopped and removed automatically.

Needed for #6

Issue #6, opened 2016-06-23
React when containers fail to start


itziakos:
This items involves the following

- detection of the failure
- reaction (UI)

**notes**:
- The design in the images from simphony-remote-docker is such that users can contribute their own images that would work with the remote app framework. That said, it is possible for an unaware user to overwrite the default ENTRYPOINT. For example, ENTRYPOINT ['sleep', '1000'] would lead to a container up and running (for 1000 sec) with a live ip and port. We may enforce the entrypoint from the docker client.
-  One way to check if the container indeed has noVNC up and running would be to do docker.top(container_name) and look for 'x11vnc' or 'websockify' ( or 'openbox-autostart'?)

kitchoi:
Close via #21

PR #16, opened 2016-06-24
Display error message at top of page when container fails to start

kitchoi:
Provide part of #6 

- When a container fails to start (timeout error, runtime error...), a user-facing error message would be shown above the list of application with a reference number
- the reference number is printed to the log, make support easier

Limitation:
- If the user starts two applications at roughly the same time and both of them fail to start, only one of the error messages would be displayed
- Issue #14 
- Issue #15

### DIFF
--- a/remoteappmanager/handlers/home_handler.py
+++ b/remoteappmanager/handlers/home_handler.py
@@ -1,6 +1,7 @@
 from collections import namedtuple
 import socket
 import os
+import uuid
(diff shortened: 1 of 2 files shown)


kitchoi:
Could do with some refactoring.  WIP for now.

kitchoi:
Attempted refactoring does not seem to me warranted, ready for review again.

Issue #18, opened 2016-06-27
Change URL id from container id to base32 encoded uuid


stefanoborini:
We found that extracting the container id from within the container is rather unreliable.
Since nothing is stopping us from personalising the url with a string passed at startup, we should go with base32 encoded uuids.

kitchoi:
Close via #19

PR #19, opened 2016-06-27
Replaced URL identifier from container id to uuid

stefanoborini:
Passes an appropriate hex string to identify the container, instead of the container ID.
Requires appropriate change in the containers.


kitchoi:
LGTM.

kitchoi:
Reference: #18

PR #21, opened 2016-06-27
Feature: UI reaction to failure to start containers

### LINKED ISSUE #14: Running containers not in cache when container is started without a public port

kitchoi:
Fixes #14 #15 and provides #6 

- `ContainerManager.containers_for_image` returns a list of Container (as before) using docker Client API instead of the cached dictionary
- Refactor the error message for `HomeHandler._actionhandler_*` to `HomeHandler.post`
- Always clean up started container in something fails in _actionhandler

### DIFF
--- a/remoteappmanager/docker/container.py
+++ b/remoteappmanager/docker/container.py
@@ -23,7 +23,7 @@ class Container(HasTraits):
     ip = Unicode()
 
     #: ...and port where the container service will be listening
-    port = Int()
+    port = Int(None, allow_none=True)
 
     @property
     def url(self):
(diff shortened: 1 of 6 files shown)


stefanoborini:
I don't get it. Did you move around stuff?

PR #22, opened 2016-06-27
Fix async docker client test on linux

kitchoi:
On linux, the `nonlinux_config` should not be used.

This also allows the test to be run on travis which has previously been skipped.


stefanoborini:
technically, it's not run because on linux DOCKER_HOST is not present, but doesn't hurt...

PR #25, opened 2016-06-28
Database data retrieval

stefanoborini:
Implements retrieval of the information from the database, and acts accordingly.

Note that this PR does not implement any security and very basic checks on the passed options.


remoteappmanager/db/orm.py
```diff
@@ -156,23 +166,60 @@ def transaction(session):
...
+
+    """
+    if user is None:
+        return False
+
+    team_ids = [team.id for team in user.teams]
+    res = session.query(Accounting) \
+        .filter(Accounting.team_id.in_(team_ids)) \
+        .filter(Accounting.application == application) \
+        .filter(Accounting.application_policy == policy).count()
```
kitchoi: `exists` would be more efficient that `count`
http://docs.sqlalchemy.org/en/latest/orm/query.html?highlight=exists#sqlalchemy.orm.query.Query.exists

tests/utils.py
```diff
@@ -61,40 +61,18 @@ def mock_docker_client_with_running_containers():
...
-        {'Command': '/sbin/init -D',
-         'Created': 1466766499,
-         'HostConfig': {'NetworkMode': 'default'},
-         'Id': 'someid',
-         'Image': 'simphony/mayavi-4.4.4:latest',
-         'ImageID': 'imageid',
-         'Labels': {'eu.simphony-project.docker.user': 'user2'},
-         'Names': ['/remoteexec-image_3Alatest_user2'],
-         'Ports': [{'IP': '0.0.0.0',
-                    'PublicIP': 34567,
```
kitchoi: I realise this should have been `PublicPort`.

tests/docker/test_container_manager.py
```diff
@@ -33,30 +35,21 @@ def test_start_stop(self):
...
-        # The output should be a list of Container
-        results = yield self.manager.containers_for_image("imageid")
-        expected = [Container(docker_id='someid',
-                              name='/remoteexec-image_3Alatest_user',
-                              image_name='simphony/mayavi-4.4.4:latest',  # noqa
-                              image_id='imageid', ip='0.0.0.0', port=None),
-                    Container(docker_id='someid',
-                              name='/remoteexec-image_3Alatest_user2',
-                              image_name='simphony/mayavi-4.4.4:latest',  # noqa
-                              image_id='imageid', ip='0.0.0.0', port=None),
```
kitchoi: Please see comment below about `PublicPort`.  Here the `port` should not be `None`.
These three Containers capture scenarios when we have unsupported containers being started.  Removing them reduces coverage as well.

tests/docker/test_container_manager.py
```diff
@@ -33,30 +35,21 @@ def test_start_stop(self):
...
-                    Container(docker_id='someid',
-                              name='/remoteexec-image_3Alatest_user3',
-                              image_name='simphony/mayavi-4.4.4:latest',  # noqa
-                              image_id='imageid', ip='', port=None)]
```
kitchoi: This container tests the scenario when the public ip is not available.

tests/docker/test_container_manager.py
```diff
@@ -67,12 +60,11 @@ def test_containers_for_image_client_api_without_user(self):
...
-        yield self.manager.containers_for_image("imageid")
+        yield self.manager.container_from_mapping_id("user", "mapping")
         call_args = self.manager.docker_client.client.containers.call_args
 
         # filters is one of the keyword argument
         self.assertIn('filters', call_args[1])
-        self.assertEqual(call_args[1]['filters']['ancestor'], "imageid")
 
     @gen_test
     def test_containers_for_image_client_api_with_user(self):
```
kitchoi: Now that user and mapping are both required parameters, only one of these two tests should be kept,

tests/docker/test_container_manager.py
kitchoi: Now I think about it, these scenarios are already tested in test_container.  It is better to remove these containers for clarity as you have done here.

remoteappmanager/handlers/home_handler.py
```diff
@@ -79,46 +54,71 @@ def post(self):
...
+
+            if len(choice) == 0:
+                raise ValueError("User is not allowed to run the application.")
+
+        _, app, policy = choice[0]
+
+        container = yield self._start_container(orm_user,
+                                                app,
+                                                policy,
+                                                mapping_id)
```
kitchoi: `docker_client.start` is not the last call in `ContainerManager._start_container`.  Not having `stop_and_remove_container` when exception occurs in `HomeHandler._start_container` leads to issue #15.  For example, if `ContainerManager.self._get_ip_and_port` fails (it does for incompatible images), it fails after the container is already started, and in this change we would not be cleaning it up.

tests/docker/test_container_manager.py
stefanoborini: Actually the test above already covers the case, so they should both go.

remoteappmanager/handlers/home_handler.py
stefanoborini: The problem is that in the exception handler then you are not guaranteed to have a container to stop. I'll add a check for that.

remoteappmanager/handlers/home_handler.py
```diff
@@ -79,46 +54,74 @@ def post(self):
...
+            container = yield self._start_container(orm_user,
+                                                    app,
+                                                    policy,
+                                                    mapping_id)
+
             yield self._wait_for_container_ready(container)
         except Exception as e:
             # Clean up, if the container is running
-            yield self._stop_and_remove_container(user_name, image_name)
+            if container is not None:
```
kitchoi: If error occurs in `self._start_container` after `docker_client.start` then `container` would be None while the container is up and runnin, i.e. still requires clean up.  Note: Can test this behaviour with an image that does not expose a port (e.g. quay.io/travisci/travis-python)

stefanoborini: Ok then, how can we clean up a container without knowing what it is...
The problem is in start_container.

remoteappmanager/db/orm.py
```diff
@@ -156,23 +167,60 @@ def transaction(session):
...
     session : Session
         The current session
     user : User
-        the orm User.
+        the orm User, or None.
 
     Returns
     -------
-    A list of tuples (orm.Application, orm.ApplicationPolicy)
+    A list of tuples (mapping_id, orm.Application, orm.ApplicationPolicy)
```
kitchoi: The console script needs updating too, we could fix that in a separate issue
(skipped part of the traceback, as it is clear from the last line what we need:)
```
$ remoteappdb --db=~/remotemanagerapp.db user list
...
  File "/mnt/hgfs/VM_shared/SimPhoNy/simphony-remote/remoteappmanager/cli/remoteappdb/__main__.py", line 92, in <listcomp>
    for app, policy in orm.apps_for_user(session, user)]
ValueError: too many values to unpack (expected 2)
```

tests/docker/test_container_manager.py
kitchoi: Not exactly.  In the test above we already mocked the return results of `docker_client.containers` to be a list of valid dictionary containing containers' info, the `filters` does not matter in that test.  We still need to check if the docker client API is utilised as expected.

remoteappmanager/handlers/home_handler.py
kitchoi: Yes.  That's what the original `HomeHandler._stop_and_remove_container` did: look for the container first and stop+remove it only if it is found.

stefanoborini: the way it was before it was passing a potentially undefined variable to stop_container.

kitchoi: I meant this in `HomeHandler`
```
-    @gen.coroutine		
 -    def _stop_and_remove_container(self, user_name, image_name):		
 -        """ Stop and remove the container associated with the given		
 -        user name and image name, if exists.		
 -        """		
 -        container_manager = self.application.container_manager		
 -        containers = yield container_manager.containers_for_image(		
 -            image_name, user_name)		
 -		
 -        # Assume only one container per image		
 -        if containers:		
 -            container_id = containers[0].docker_id		
 -            yield container_manager.stop_and_remove_container(container_id)
```

kitchoi: That said, we can surely do things differently in this regard.

remoteappmanager/handlers/home_handler.py
kitchoi: Would it be possible that in ContainerManager._start_container we make sure `docker_client.start` is the last thing to do?  So if it fails, we can be certain that the container is not started.

stefanoborini: We can't get the port without starting the container. Plus, we created it anyway, so we need to delete it. I think that the main problem is that we should do the rollback inside start_container.

kitchoi: Alternatively we could do `return yield container_manager.container_from_mapping_id(orm_user.name, mapping_id)` inside `ContainerManager._start_container`?  In this case we would want to add an (`all=True`) flag in `container_from_mapping_id` to get a stopped container.

remoteappmanager/docker/container_manager.py
```diff
@@ -94,44 +90,29 @@ def stop_and_remove_container(self, container_id):
...
-            The image id or name
-
-        Optional parameters
-        -------------------
-        user_name : str
-            Name of the user who started the container
+        user_name: str
+            The username
+        mapping_id : str
+            The unique id to identify the container
```
kitchoi: mapping_id is unique for a given user.  However it is not necessarily unique among different users, we should clarify that.

kitchoi: In fact, technically we have not restricted `mapping_id` to be unique because `start_container` can be called multiple times (who knows how) with the same `mapping_id`.

kitchoi: To clarify, the docstring reads to me that mapping_id is a 'unique identifier for one container', which is not necessarily true for a given user as I explained above.

remoteappmanager/docker/container_manager.py
```diff
@@ -94,44 +90,29 @@ def stop_and_remove_container(self, container_id):
...
+        }
 
         containers = yield self.docker_client.containers(filters=filters)
-        return [Container.from_docker_dict(container)
-                for container in containers
-                # Require further filtering as ancestor include grandparents
-                if (container.get('Image') == image_id_or_name or
-                    container.get('ImageID') == image_id_or_name)]
+        if len(containers) != 0:
+            return Container.from_docker_dict(containers[0])
```
kitchoi: Following the above comments regarding the uniqueness of mapping_id, there can be multiple containers for one mapping_id and one user, which would be in fact desired when multiple instances of the same application+policy is allowed.  In that case, containers could be a list with more than one item.

remoteappmanager/docker/container_manager.py
```diff
@@ -165,7 +158,7 @@ def _start_container(self, user_name, image_name, volumes):
         # build the dictionary of keyword arguments for create_container
         container_name = _generate_container_name("remoteexec",
                                                   user_name,
-                                                  image_name)
+                                                  mapping_id)
```
kitchoi: Ah this does restrict the username+mapping_id to be a unique identifier for one container.  But as far as I understand, we allow multiple containers per user+application+policy, no?  This configuration will raise an error when someone wants to create a second instance of the same application+policy.

remoteappmanager/handlers/home_handler.py
kitchoi: Please see my comment below regarding `container_from_mapping_id`.  Perhaps we could do a simple docker_client.containers(name={we_already_know_the_name}) and pass it to `ContainerManager.from_docker_dict`, which will take care of missing ip and port without raising an error.

kitchoi:
Review done.  There are number of key issues above that we probably want to discuss further.

remoteappmanager/handlers/home_handler.py
kitchoi: Actually, taking a step back, if we don't mind changing the interface of the method, we should stick with either having side-effect or having return value, but not both; i.e. start_container should return None.

remoteappmanager/docker/container_manager.py
stefanoborini: It's based on a uuid so it's unique within the constraints of uuid collisions.

stefanoborini: Ah sorry, I misunderstood your point.

remoteappmanager/docker/container_manager.py
stefanoborini: I don't think we are considering the case at all. The user is not allowed to start a container with the same configuration multiple times.

remoteappmanager/docker/container_manager.py
stefanoborini: Not really. We support only one instance (or at least that was the initial idea. I don't know if something changed in this vision in the meantime)

stefanoborini:
@kitchoi Addressed some of the issue, please re-review. I am going to go through the diff once more. The API has changed a bit, so I might have to introduce a few more tests.

remoteappmanager/docker/container_manager.py
kitchoi: It seems that there is an inconsistency of my memory and your memory of what the mapping id is used for.  According to my problematic memory, the mapping id is a unique identifier for application+policy (or just policy?) so that, for example if the policy says only 10 instances is allowed, we make sure there are no more than 10 containers with this mapping id.  Nothing is said about how these instances are shared among users.

stefanoborini: Then we have two completely different levels of accounting.

1. the number of different applications an individual user can start.
2. the number of containers a team of users can spawn of the same application.

These are different. My point is that user A should be allowed to run one container of application B with configuration C. He can however run a second container of application B with configuration C2, or a second container of application B2 with configuration C. Each of these containers will have different mapping_id.

remoteappmanager/handlers/home_handler.py
```diff
@@ -79,46 +54,74 @@ def post(self):
...
-            # the container url to it.
-            self.application.reverse_proxy_add_container(container)
 
-            # Redirect the user
-            url = self.application.container_url_abspath(container)
-            self.log.info('Redirecting to ' + url)
-            self.redirect(url)
+        # The server is up and running. Now contact the proxy and add
+        # the container url to it.
+        self.application.reverse_proxy_add_container(container)
```
kitchoi: oops, my bad, yield is needed here.

stefanoborini: Adding it. one sec.

remoteappmanager/cli/remoteappdb/__main__.py
```diff
@@ -89,7 +89,7 @@ def list(ctx):
                                          policy.volume_source,
                                          policy.volume_target,
                                          policy.volume_mode)
-                    for app, policy in orm.apps_for_user(session, user)]
```
itziakos: This comment is related to the functiion docstring. Please update it since it refers to users instead of images.

remoteappmanager/db/orm.py
```diff
@@ -11,28 +11,35 @@
...
+
 class UserTeam(Base):
     """ The user (n <-> n) team association table """
     __tablename__ = "user_team"
     user_id = Column(Integer, ForeignKey('user.id'), primary_key=True)
     team_id = Column(Integer, ForeignKey('team.id'), primary_key=True)
 
 
-class User(Base):
+class User(IdMixin, Base):
```
itziakos: Why not have a new base class subclass of `Base` for all those table entries that have the `id` attribute. Is the use of mixins idiomatic to how orms are setup?

stefanoborini: Because then sqlalchemy pretends to have a table for the parent class, and working around it is less smooth than using a mixin like that (which is according to sqlalchemy docs)

remoteappmanager/cli/remoteappdb/__main__.py
stefanoborini: which comment?

remoteappmanager/docker/container_manager.py
```diff
@@ -62,7 +54,11 @@ def start_container(self, user_name, image_name, volumes=None):
             The name of the user
         image_name: string
             A string identifying the image name.
-        volumes: dict
+        mapping_id: str
+            A generic id used to recognize the container. It is not unique
```
itziakos: I would just state that this id is expected to be unique (and persistent) for a specific combination of docker image (i.e. application) and setup (i.e. configuration).

kitchoi:
I've finished reviewing.  Other than the fact I need some clarification for the usage of the mapping id, the rest looks good to me.

remoteappmanager/docker/container_manager.py
itziakos: To clarify:

**application**: The docker image to use 
**policy**: The container setup for connecting volumes and the execution constrains 

The mapping id is unique for each combination of (application, policy) tuple. The id is persistent across runs of the single user app and jupyter hub server.

kitchoi: The key question here is, can a user have multiple container instances for the same (application, policy) tuple?

itziakos: My current understanding of the features that I want for the remote app is that we we should not allow multiple instances with the same application,policy from the same user

remoteappmanager/cli/remoteappdb/__main__.py
itziakos: please ignore my comment

remoteappmanager/handlers/home_handler.py
kitchoi: The cleanup issue is resolved.

Issue #5, opened 2016-06-23
Better testing for `HomeHandler`


itziakos:
We need a better testing infra for tornado web handlers.

stefanoborini:
I envision two possible testing strategies:

1. Using the tornado services.
2. Doing actual requests.

We will have to mock the external services though.

stefanoborini:
We encounter a tornado bug when dealing with this issue. https://github.com/tornadoweb/tornado/issues/1573 

I'll try to work around it a bit, but seems hard to do so without hacking tornado itself.

stefanoborini:
Ready in pull request #48, although I recommend that we expand this approach as we will need to test other handlers (e.g. the REST api handler, or new methods).

I suggest to

- introduce fully functional mock objects for all the services that we use, with a clear "mock infrastructure" that returns something sensible and functional.
- Introduce the Form object to factor out validation of form submissions, so that POST handling becomes a bit cleaner and we can focus on testing the Form object, instead of the handler.

PR #27, opened 2016-06-28
HomeHandler.get and post requires authentication

### LINKED ISSUE #24: Home page of remoteappmanager does not require authentication
Suppose a remoteappmanager is spawned for `user=username` and the reverse proxy is set for `/user/username`, anyone can access `https://hub_server_url/user/username` without logging in.

1. Log in as username
2. Open an incognito browser, go directly to https://hub_server_url/user/username

kitchoi:
Fix #24 

Should be revisited after #25

### DIFF
--- a/remoteappmanager/handlers/base_handler.py
+++ b/remoteappmanager/handlers/base_handler.py
@@ -1,5 +1,9 @@
+import requests
+from urllib.parse import urljoin, quote
+
 from tornado import web
-from urllib.parse import urljoin
+from tornado.httpclient import HTTPError
+from notebook.utils import url_path_join
 
 from remoteappmanager.logging.logging_mixin import LoggingMixin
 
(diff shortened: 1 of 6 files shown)


kitchoi:
Tests pass, but handlers are not yet tested (issue #5 ) so there is a bit reduction in coverage...

stefanoborini:
@kitchoi We need to address coverage in more detail as we proceed, but for now I think it's fine. We need to get a baseline and build from there.

kitchoi:
@sbo Please review.

remoteappmanager/handlers/base_handler.py
```diff
@@ -17,8 +28,35 @@ def render(self, template_name, **kwargs):
...
             user=self.current_user,
             base_url=self.application.config.base_url,
-            logout_url=urljoin(self.application.config.hub_prefix, "logout")
-        )
+            logout_url=urljoin(self.application.config.hub_prefix, "logout"))
 
         args.update(kwargs)
         super(BaseHandler, self).render(template_name, **args)
+
+    def verify_token(self, cookie_name, encrypted_cookie):
```
stefanoborini: My only concern is that this code is not asyncronous, meaning that we assume the request operation to be quick. This code is from jupyterhub, and they do the same, but it's a potential concern for us.

I am going to merge this, but we might have to rework this routine with a threadpool (like in async docker client) and issue the request asynchronously.

stefanoborini:
Not concerned about the codecov. We are going to aim for a baseline at 0.2 release, and move up from there.

kitchoi:
I agree that we need to revisit the async issue.

Issue #4, opened 2016-06-23
Attach workspace to virtual users


itziakos:
The task involves the following:

- hub server starts as hub-server
- Authentication is custom
- Single user manager app starts as hub-server
- The single user manager creates if necessary a dedicated folder for a new user.
- The single user manager attaches the dedicated folder to the new container as /worspace
- Provide support for upload/download of the files in the workspace (probably at the level of the single user manager).

kitchoi:
Just wanted to clarify, the hub server may or may not be run as root?

itziakos:
Yes, this setup should not require the Jupyter Hub server to be started as root.

itziakos:
please note that 
> Provide support for upload/download of the files in the workspace (probably at the level of the single user manager).

if #9 which is designated for milestone 0.4.0

kitchoi:
@itziakos Thanks!  I did not notice that #9 is in 0.4.0...

stefanoborini:
@kitchoi What's the status of this one? Finished or there's still work to do? should it be moved to 0.4.0?

kitchoi:
@stefanoborini @itziakos The first 5 bullet points in this task is completed via #36.
The last bullet point regarding upload/download, as far as I understood, is in milestone 0.4.0, so I am not working on it at the moment.

stefanoborini:
@kitchoi Ok, then I close this.

PR #36, opened 2016-07-01
Feature local Virtual user spawner

kitchoi:
Provides part of #4 (lacking the upload/download facility)

~~Still need to add tests.  Putting this up to solicit feedbacks.~~

Features:
- Added `VirtualUserSpawner` so that all processes are owned by the user who started the hub server
- Added a WorldAuthenticator that authenticates anyone without password
- If `VirtualUserSpawner.workspace_dir` is set, a temporary directory is created for the user when the single user app is spawned, and is cleaned up when the app is stopped;  the directory is then mounted as workspace
- One database is shared among all users


stefanoborini:
Postponing for 0.3

kitchoi:
After restarting the singleuser app, mounted /workspace in **started** containers would no longer be available, since the VirtualUserSpawner.stop removes the temporary workspace on the host machine.  

You get 'No such file or directory' inside the container.  Fortunately this time docker does not try to create the nonexisting directory.

Either we state this as a limitation, or we clean up the started containers whenever the singleuser app exits, or we keep a persistent directory for the user workspace instead of a temporary one.
Thought?

stefanoborini:
Reference issue #4

jupyterhub/jupyterhub_config.py
```diff
@@ -1,10 +1,42 @@
...
 from jupyter_client.localinterfaces import public_ips
 
 c = get_config()  # noqa
 
 c.JupyterHub.ssl_key = 'test.key'
 c.JupyterHub.ssl_cert = 'test.crt'
-c.JupyterHub.spawner_class = 'remoteappmanager.spawner.Spawner'
 
 c.JupyterHub.hub_ip = public_ips()[0]
+
```
stefanoborini: I think that if we want to add a different authenticator, it should be in our code, not in the config file.

jupyterhub/remoteappmanager_config.py
```diff
@@ -29,7 +29,8 @@
...
 # 1. this database is shared among all instances of the remoteappmanager.
-# 2. if a relative path is specified, it will be relative to the working
-#    directory specified at startup, if specified, otherwise to the cwd
-#    at startup. Note that jupyterhub spawners spawn in the home directory.
-db_url = "sqlite:///"+os.path.expanduser("~/remoteappmanager.db")
+# 2. When running with system-user mode, the jupyterhub spawners spawn in
+#    the user's home directory; when running in virtual-user mode,
+#    the current directory is the directory where jupyterhub is started
+# 3. '~' would be evaluated as the spawned user's home directory
+db_url = "sqlite:///"+os.path.abspath("./remoteappmanager.db")
```
stefanoborini: I am not sure this is the case. I suspect that this will just give the absolute path relative to the cwd, which will still be the user home directory. chdir is done before the spawn.

remoteappmanager/handlers/home_handler.py
```diff
@@ -265,8 +265,12 @@ def _start_container(self, orm_user, app, policy, mapping_id):
         volumes = {}
 
         if mount_home:
-            home_path = os.path.expanduser('~'+user_name)
-            volumes[home_path] = {'bind': '/workspace', 'mode': 'rw'}
+            home_path = os.environ.get('HOME')
```
stefanoborini: even if you are running with virtual users, HOME will be defined (and it's the jupyterhub user). This will mount the jupyterhub home directory for all containers that have mount_home specified.

jupyterhub/jupyterhub_config.py
kitchoi: Yes I agree.
The WorldAuthenticator is only a place holder for replacing the PAMAuthenticator.  We don't know what Authenticator we want yet.

remoteappmanager/spawner.py
```diff
@@ -39,3 +43,129 @@ def _config_file_path_default(self):
...
+        state = super().get_state()
+        if self._virtual_workspace:
+            state['virtual_workspace'] = self._virtual_workspace
+        return state
+
+    def clear_state(self):
+        super().clear_state()
+        self._virtual_workspace = ''
+
+    def get_args(self):
```
stefanoborini: This is common with the other class

remoteappmanager/spawner.py
```diff
@@ -39,3 +43,129 @@ def _config_file_path_default(self):
...
+    def start(self):
+        """ Start the process and create the virtual user's
+        temporary home directory if `workspace_dir` is set
+        """
+        # Create the temporary directory as the user's workspace
+        if self.workspace_dir and not self._virtual_workspace:
+            try:
+                self._virtual_workspace = tempfile.mkdtemp(
+                    dir=self.workspace_dir)
+            except Exception as exception:
```
stefanoborini: you don't need to do as. Just use self.log.exception(). It will add the content of the current exception automatically.

remoteappmanager/spawner.py
```diff
@@ -39,3 +43,129 @@ def _config_file_path_default(self):
...
+                               'for %s in %s  with error: %s',
+                               self.user.name, self.workspace_dir, exception)
+            else:
+                self.log.info("Created %s's temporary workspace in: %s",
+                              self.user.name, self._virtual_workspace)
+
+        # Make sure we clean up in case `start` fails
+        try:
+            yield super().start()
+        except Exception as exception:
```
stefanoborini: same, just raise.

remoteappmanager/spawner.py
```diff
@@ -39,3 +43,129 @@ def _config_file_path_default(self):
...
+
+    def _clean_up_workspace_dir(self):
+        """ Clean up the virtual user's temporary directory, if exists
+        """
+        if not self._virtual_workspace:
+            return
+
+        # Clean up the directory
+        # Make sure the temporary directory is not /, ./ or ../
+        if self._virtual_workspace.strip('/') in ('', '.', '..'):
```
stefanoborini: ./.

tests/spawner/test_spawner.py
```diff
@@ -1,54 +1,84 @@
...
 from jupyterhub import orm
-from remoteappmanager.spawner import Spawner
+
+from remoteappmanager.spawner import Spawner, VirtualUserSpawner
 from tests import fixtures
 
 
+def start_spawner(spawner):
+    io_loop = IOLoop.current()
+    io_loop.run_sync(spawner.start)
```
stefanoborini: This will run it synchronously and exit the event loop. There's no point in waiting after this. Maybe you want to run later?

remoteappmanager/handlers/home_handler.py
kitchoi: But HOME is set by the VirtualUserSpawner.

remoteappmanager/spawner.py
kitchoi: Now we only have two classes.  I would wait until we have more classes before refactoring and subclassing a base spawner of our own.

jupyterhub/remoteappmanager_config.py
kitchoi: chdir is neutralised by the `VirtualUserSpawner.make_preexec_fn`
With the VirtualUserSpawner, the cwd is where jupyterhub is started.
(Do I want to copy most of the code of LocalProcessSpawner  and subclass Spawner or just overload a handful of its methods? hmmm)

remoteappmanager/handlers/home_handler.py
stefanoborini: Let me check. If it's the shell that sets it, then we are in the clear, provided jupyterhub is invoking the process without a shell.

stefanoborini: According to http://superuser.com/questions/271925/where-is-the-home-environment-variable-set it's exported by the login process. Which means that it's always present. Question is if what you are doing to overwrite it also exports it to subprocesses... let me check.

tests/spawner/test_spawner.py
kitchoi: removing `time.sleep(1)` gives me these failures in tests:
```
======================================================================
FAIL: test_spawner_start_and_stop (tests.spawner.test_spawner.TestSpawner)
----------------------------------------------------------------------
  File "/mnt/hgfs/VM_shared/SimPhoNy/simphony-remote/tests/spawner/test_spawner.py", line 134, in test_spawner_start_and_stop
    self.assertEqual(status, 1)
AssertionError: -2 != 1

======================================================================
FAIL: test_spawner_with_workspace_dir (tests.spawner.test_spawner.TestVirtualUserSpawner)
----------------------------------------------------------------------
  File "/mnt/hgfs/VM_shared/SimPhoNy/simphony-remote/tests/spawner/test_spawner.py", line 183, in test_spawner_with_workspace_dir
    self.assertEqual(status, 1)
AssertionError: -2 != 1

======================================================================
FAIL: test_spawner_without_workspace_dir (tests.spawner.test_spawner.TestVirtualUserSpawner)
----------------------------------------------------------------------
  File "/mnt/hgfs/VM_shared/SimPhoNy/simphony-remote/tests/spawner/test_spawner.py", line 162, in test_spawner_without_workspace_dir
    self.assertEqual(status, 1)
AssertionError: -2 != 1

======================================================================
FAIL: test_start_if_workspace_path_not_exists (tests.spawner.test_spawner.TestVirtualUserSpawner)
----------------------------------------------------------------------
  File "/mnt/hgfs/VM_shared/SimPhoNy/simphony-remote/tests/spawner/test_spawner.py", line 242, in test_start_if_workspace_path_not_exists
    self.assertEqual(status, 1)
AssertionError: -2 != 1

----------------------------------------------------------------------
Ran 50 tests in 10.333s
```

jupyterhub/remoteappmanager_config.py
stefanoborini: Don't know... we can do a base class, but then it's probably quite useless in itself... 
on the other hand, if we add one --option or an envvar, I don't want to run the risk that we update in one and not the other...

tests/spawner/test_spawner.py
stefanoborini: Ok, there's something weird going on here. according to the docs, run_sync starts the event loop, runs the future inside the single iteration of the loop, and exits when the future is completed. There's no apparent reason why the sleep would have an effect.

jupyterhub/remoteappmanager_config.py
kitchoi: > if we add one --option or an envvar, I don't want to run the risk that we update in one and not the other...

That's true.  Instead of having a base class, have a base test case?

tests/spawner/test_spawner.py
kitchoi: I notice that tests in jupyterhub does the same thing (`time.sleep`).  Perhaps it has to do with delay in the singleruserapp starting its own event loop?

remoteappmanager/handlers/home_handler.py
kitchoi: This thread is clear. (See https://github.com/simphony/simphony-remote/pull/36#discussion_r69477616)

kitchoi:
To summarise, this is only issue from the above:
https://github.com/simphony/simphony-remote/pull/36#discussion_r69478359:
>> if we add one --option or an envvar, I don't want to run the risk that we update in one and not the other...
>
> That's true.  Instead of having a base class, have a base test case?

Thought on this?
And did I miss anything else?

jupyterhub/jupyterhub_config.py
stefanoborini: You can move it... It's ok if it's part of it. It's a good utility and moving it keeps the config clean.

jupyterhub/remoteappmanager_config.py
stefanoborini: It's more than the testcase. It's a potential problem in development.

tests/spawner/test_spawner.py
stefanoborini: Ok, if it's because of a race condition with the shell spawn, then keep the sleep.

jupyterhub/remoteappmanager_config.py
kitchoi: I will put a TODO comment (as a reminder for updating the options as well), and let's open issue for it.

jupyterhub/jupyterhub_config.py
kitchoi: Moved WorldAuthenticator back to the code.  Codecov is going to complain...

kitchoi:
Good to merge?

stefanoborini:
:+1:

PR #44, opened 2016-07-04
Remove teams

stefanoborini:
Removed the Teams table, and the associated services and tests.


stefanoborini:
There's a problem with click applications. It's rather hard to get them to pass the patch coverage, because even the click testing framework does an invocation test. As such, coverage can't see the execution.

kitchoi:
LGTM

Issue #33, opened 2016-06-30
Abstract database interface to allow for other data providers (e.g. csv file)


stefanoborini:
We need to abstract the database interface so that we can use a csv file instead of the database. ORM objects need to be duck typed, so that we can work with different (non-ORM) classes transparently.

kitchoi:
Close via #47

PR #47, opened 2016-07-04
Feature database interface

kitchoi:
Provides for #33


tests/db/abc_test_interfaces.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def test_get_apps_for_user(self):
+        """ Test get_appgs_for_user returns an iterable of ApplicationConfig
+        """
+        database = self.create_database()
+
+        for username, expected_configs in self.user_config_mappings.items():
+            actual_configs = database.get_apps_for_user(username)
+            self.assertSetEqual(set(actual_configs), set(expected_configs))
+
```
kitchoi: Shall we check that the items in the `actual_config` are instance of `ApplicationConfig`? Or are we lenient enough that any types providing the same sets of trait names as property/attribute is good enough If it is the latter, ApplicationConfig might well be a namedtuple.

stefanoborini: `ApplicationConfig` is a bit deceiving as name. We currently have another object with that name (soon to be split) but still remains deceiving because it is not the config of the remoteappmanager. I understand that you are technically right, but it's deceiving at first glance.

kitchoi: Any suggestion what to call? *:naming-things-is-hard:*

stefanoborini: I would not use it. If I understand correctly from your design, you plan to abstract away the orm objects. I would not. I would keep them as interfaces, have the orm classes implement the interface, and use the same interface for plain old python objects that honor the interface even when sourced from csv.

I don't want to abstract too much. Every layer adds code.

kitchoi: Are you referring to just `ApplicationConfig`(whatever it may be called otherwise)?

kitchoi:
@stefanoborini you don't need to change anything with the ORM.  You just need to provide the Database interfaces `get_user_by_name` and `get_apps_for_user`, with the latter returns `ApplicationConfig` (or whatever it should be called).  This is very basic and contains all the web application/docker cares about.

stefanoborini:
@kitchoi what I mean is that you are basically hiding the implementation behind an additional layer of abstraction (the ApplicationConfig). I would expect that we have to copy all the information from the orm objects to the ApplicationConfig object, and I find it a bit annoying...

kitchoi:
I first pictured using ApplicationConfig as an immutable but I did not implemented it as one!  It might be better to just use a namedtuple and forget about type checking and auto init.  The current implementation of ApplicationConfig is mutable and it should not be hashable either.  I am going to modify `__hash__`

kitchoi:
Now I miss hash.  Mutable or immutable?  Which should we pick.

kitchoi:
I have a branch here that demos how to use the Database interface with the current sqlalchemy database:
https://github.com/simphony/simphony-remote/tree/demo-database-interface

remoteappmanager/db/interfaces.py
```diff
@@ -0,0 +1,81 @@
...
+        return "<{cls}({attrs})>".format(
+            cls=self.__class__.__name__,
+            attrs=", ".join("{0}={1!r}".format(
+                trait_name, getattr(self, trait_name))
+                            for trait_name in self.trait_names()
+                        )
+        )
+
+
+class ABCDatabase(metaclass=ABCMeta):
```
itziakos: I would call that `Accounting`

remoteappmanager/db/interfaces.py
```diff
@@ -0,0 +1,81 @@
...
+
+    #: Target path (on the container) for the application specific data
+    #: volume
+    volume_target = Unicode(allow_none=True)
+
+    #: Read-write access mode for the application specific data volume
+    #: (For docker, it is either "rw": read-write or "ro": read-only
+    volume_mode = Unicode(allow_none=True)
+
+    def __eq__(self, other):
```
itziakos: Why do we need __eq__?

remoteappmanager/db/interfaces.py
```diff
@@ -0,0 +1,81 @@
...
+        if not isinstance(other, ApplicationConfig):
+            return False
+
+        for trait_name in self.trait_names():
+            if getattr(self, trait_name) != getattr(other, trait_name):
+                return False
+
+        return True
+
+    def __hash__(self):
```
itziakos: `__hash__ = None` should work for python 3 also correct?

remoteappmanager/db/interfaces.py
kitchoi: The alternative would be to implement `assertAccountingsEqual` in tests.

remoteappmanager/db/interfaces.py
kitchoi: Yes.  Didn't know that!  Thanks.

remoteappmanager/db/interfaces.py
kitchoi: ABCAccounting? IAccounting? or just Accounting?

tests/db/abc_test_interfaces.py
```diff
@@ -0,0 +1,45 @@
...
+        """ Test ABCDatabase.get_user_by_name """
+
+    def test_get_apps_for_user(self):
+        """ Test get_appgs_for_user returns an iterable of ApplicationConfig
+        """
+        database = self.create_database()
+
+        for username, expected_configs in self.user_config_mappings.items():
+            actual_configs = database.get_apps_for_user(username)
+            self.assertSetEqual(set(actual_configs), set(expected_configs))
```
kitchoi: now that the config is not hashable, this is where the test fails.

remoteappmanager/db/interfaces.py
itziakos: if it is only for tests then I would prefer to use https://docs.python.org/3/library/unittest.html?highlight=unittest#unittest.TestCase.addTypeEqualityFunc

This will allow to treat the objects immutable

remoteappmanager/db/interfaces.py
```diff
@@ -0,0 +1,81 @@
+from abc import ABCMeta, abstractmethod
+
+from traitlets import HasTraits, Unicode, Bool
+
+
+class ApplicationConfig(HasTraits):
```
itziakos: I do not think that there is any advantage in having these as HasTraits class. I would expect this to but just a simple able class.

remoteappmanager/db/interfaces.py
itziakos: `ABCAccounting`

itziakos:
@kitchoi ApplicationConfig, Accounting and User should be ABC classes for which different implementation register their objects (https://docs.python.org/3/library/abc.html?highlight=abc#abc.ABCMeta.register). They should not be traitlets. 

The abc description should define the attributes in an `__init__` method (no `abstractproperty`) and document them according to the autodoc rules.
```
def __init__(self, name):
      #: The name of the user
      self.name = name

However I thought that there was and `Application` and `Policy` objects instead of ApplicationConfig. I would prefer to keep them separate for the moment. 

To summarize:
Te ABC are there to describe and document the desired api and object. Concrete implementations can subclass and reuse but they do not have to (they can use `register`).  In reality we use duck typing but in a more formal way.

**Comment by kitchoi** [comment]
Thanks @itziakos 
I am fine with separating the `Application` and `ApplicationPolicy` for now.
For `ABCAccounting`, I still think it should be `ABCDatabase` for the fact that those query methods are general to the database.  The `orm.Database` should subclass it, not `orm.Accounting`.  In practice, one can call these methods inside the remoteappmanager tornado application by calling `application.db.get_user_from_name` or `application.db.get_apps_for_user`.

File: remoteappmanager/db/interfaces.py (hunk 6 above)
Comment (kitchoi) [inline remoteappmanager/db/interfaces.py]: There is one advantage which is to be able to loop over the trait names when checking for 'equal'.  Without it we need to make sure we update the tests when we add attributes.  Alternatives?

**Comment by itziakos** [comment]
> For  ABCAccounting , I still think it should be  ABCDatabase  for the fact that those query methods are general to the database. The  orm.Database  should subclass it, not  orm.Accounting 

I think that there is a confusion between the ABCAcounting and the database accounting table. They are not the same. The description we provide here does not depend on the database. For the application the ABCAccounting object provide a set of functions to get info about the users and policy.

**Comment by kitchoi** [comment]
> I think that there is a confusion between the ABCAcounting and the database accounting table. They are not the same. The description we provide here does not depend on the database. For the application the ABCAccounting object provide a set of functions to get info about the users and policy.

I agree and thus I understood `ABCAccouting` to be used as `remoteappmanager.Application.db = Instance(ABCAccounting)`.  However the name `ABCAccounting` would suggest it being used by the `orm.Accounting`, rather than the `orm.Database`.  Am I making sense?

**Comment by kitchoi** [comment]
To clarify, I mainly oppose the name...

File: remoteappmanager/db/interfaces.py (hunk 6 above)
Comment (itziakos) [inline remoteappmanager/db/interfaces.py]: If you want a quick function for comparing you could even compare the `json` strings of the two objects. But a simple function comparing the attributes should be fine. They are not many of them and the interface is not going to change so often.

**Comment by itziakos** [comment]
> To clarify, I mainly oppose the name...

ok, so ABCDatabase is also bad because the object is not describing a database. I used the word `Accounting` because it points to the Accounting management from the point of the application. Based on that it is unfortunate that there is already a table named Accounting in the database (however this is only an internal schema item).

I would prefer `ABCAccounting`  from `ABCDatabase`. But I am ok with something else if we can agree on a name. Lets take this offline

File: remoteappmanager/db/interfaces.py (hunk 6 above)
Comment (kitchoi) [inline remoteappmanager/db/interfaces.py]: json complains that the object cannot be serialised.  You meant this right?:
```
import json
json.dumps(instance, sys.stdout)
```

**Comment by kitchoi** [comment]
@stefanoborini @itziakos 
Here is what I understood so far for this abstraction:

From the point of view of the sqlalchemy database, the ABCApplication, ABCApplicationPolicy are merely there to tell you what to duck type.  orm.Application and orm.ApplicationPolicy do not subclass them, nor is it useful to register them as virtual subclass.   However, a class that implements the interface of `ABCAccounting` is required for the sqlalchemy support.

For other database interface (e.g. csv), ABCApplication and ABCApplicationPolicy should be subclassed when possible (though if you want to use traitlets classes, you can't subclass these ABCs either).

File: remoteappmanager/db/interfaces.py [hunk 7]
```diff
@@ -0,0 +1,90 @@
...
+        user_name : str
+
+        Returns
+        -------
+        user : User
+            a User object that the Database understands
+        """
+
+    @abstractmethod
+    def get_apps_for_user(self, user_name):
```
kitchoi: instead of just `user_name (str)`, maybe `user_name(str)` or `user(User)` (results of `get_user_by_name`)

kitchoi: Perhaps just `user(User)` (result of `get_user_by_name`, which can be anything the database understands)

tests/db/abc_test_interfaces.py
```diff
@@ -0,0 +1,94 @@
+from abc import abstractmethod, ABCMeta
+from collections import Iterable
+import inspect as _inspect
+
+from remoteappmanager.db.interfaces import ABCApplication, ABCApplicationPolicy
+
+
+class ABCTestDatabaseInterface(metaclass=ABCMeta):
+
+    def assertApplicationEqual(self, app1, app2, msg=None):
```
stefanoborini: why camelcase?

stefanoborini: Ignore... Makes sense.

PR #46, opened 2016-07-04
Split of configuration options objects

stefanoborini:

PR #48, opened 2016-07-04
Home handling testing

stefanoborini:


stefanoborini:
In progress, and needs previous merge of pull request #46

stefanoborini:
Leaving this PR for later. I need to play with the database, monkey patching here and there, and it makes sense that we wait until the database abstraction is settled.

stefanoborini:
After the introduction of PR #52 this will break again, so it's a better plan to merge that one first.

stefanoborini:
@kitchoi ready for review

tests/handlers/test_home_handler.py
```diff
@@ -0,0 +1,206 @@
...
+    def test_home(self):
+        res = self.fetch("/user/username/",
+                         headers={
+                             "Cookie": "jupyter-hub-token-username=foo"
+                         }
+                         )
+
+        self.assertEqual(res.code, 200)
+        self.assertIn("Available Applications", str(res.body))
+
```
kitchoi: Looks like we should have a test for failing to verify token.  e.g.
```
def test_home_unauthorised(self):
    self._app.hub.verify_token.return_value = False
    ...
```

kitchoi: ah no, verify_token raises an error if not authorised...

kitchoi: `self._app.hub.verify_token.side_effect = HttpError(mock.Mock(status=403), 'Auth token may have expired')`

kitchoi: We can also add more testes later.  This PR provides a framework for that.

tests/temp_mixin.py
```diff
@@ -9,14 +9,14 @@ class TempMixin:
     """
 
     def setUp(self):
-        super().setUp()
         self.tempdir = tempfile.mkdtemp()
+        super().setUp()
```
kitchoi: If `super().setUp()` fails here, the temporary directory is created but then the `tearDown` would not be called.  Instead, have a `remove_tempdir` method and `self.addCleanup(self.remove_tempdir)` right after `mkdtemp` would ensure that the temporary directory is removed.

kitchoi:
Thanks @stefanoborini 
Except for one comment on TestCase.tearDown, the rest looks good to merge!

Issue #28, opened 2016-06-29
Add more functionality to the remoteappdb script


stefanoborini:
We need a few more functionality to the program. Specifically:

- Better formatting of the produced tables.
- Ability to remove users, teams, users from teams, applications.

stefanoborini:
Will start working on this tomorrow, but I need to have the other PRs merged first.

PR #50, opened 2016-07-05
Enhanced remoteappdb

stefanoborini:
Provides an enhanced version of the remoteappdb that allows removal of users and apps, and better error report.

### DIFF
--- a/remoteappmanager/cli/remoteappdb/__main__.py
+++ b/remoteappmanager/cli/remoteappdb/__main__.py
@@ -113,8 +113,8 @@ def remove(ctx, user):
 @click.option('--no-decoration', is_flag=True,
               help="Disable table decorations")
 @click.option('--show-apps', is_flag=True,
-              help="For each user, shows the "
-                   "applications he is allowed to run")
+              help="Shows the applications each user "
+                   "is allowed to run")
 @click.pass_context
 def list(ctx, no_decoration, show_apps):
     """Show a list of the available users."""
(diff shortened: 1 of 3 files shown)


kitchoi:
Reference #28

kitchoi:
Testing the utility manually, I notice there is a retrogression in assigning the volume_mode.  The `volume_mode` is always `None` (not just displayed as None but actually None in the database).

```
(simremote)kit@kit-virtual-machine:simphony-remote$ remoteappdb --db=testing.db app grant simphonyproject/empty-ubuntu tester --allow-home --volume /appdata/image_name/common:/appdata:ro
(simremote)kit@kit-virtual-machine:simphony-remote$ remoteappdb --db=testing.db user list --show-apps
ID    Name    App                             Home    View    Common  Vol. Source                 Vol. Target      Vol. Mode
----  ------  ----------------------------  ------  ------  --------  --------------------------  -------------  -----------
1     kit     simphonyproject/empty-ubuntu       1       1         0
              simphonyproject/empty-ubuntu       1       0         1  /appdata/image_name/common  /appdata
2     tester  simphonyproject/empty-ubuntu       1       0         1  /appdata/image_name/common  /appdata
```

kitchoi:
Regarding cascading delete, this might be useful? (if you've tried and it doesn't work, please ignore)
http://docs.sqlalchemy.org/en/latest/orm/cascades.html#delete

kitchoi:
Regarding the cascading delete reference, that also requires changes in the orm itself. We might want to do that in a separate PR if we want to go for it.

stefanoborini:
We can't do cascading delete. sqlite does not support it.

kitchoi:
> We can't do cascading delete. sqlite does not support it.

Okay.  Thanks for checking.

stefanoborini:
I need to revise my statement. sqlite does support it, but it needs to be enabled, and it's not by default. That's why it didn't work when I tried. I'll have to study a bit on this regard and as you said, change the model.

kitchoi:
LGTM

PR #51, opened 2016-07-06
Feature adapt database interface

kitchoi:
Depends on #47 

the `remoteappmanager.db` is now an instance of `orm.AccountingInterface` which subclass `ABCAccounting`

- [x] Add test for `orm.AccountingInterface`

Note: AccountingInterface <--- I am not sold on this name :(

### DIFF
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -2,7 +2,6 @@ import os
 from urllib import parse
 
 from traitlets import Instance
-from sqlalchemy.orm.exc import MultipleResultsFound
 from tornado import web, gen
 import tornado.ioloop
 from jinja2 import Environment, FileSystemLoader
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -9,6 +9,7 @@ from jinja2 import Environment, FileSystemLoader
 from jupyterhub import orm as jupyterhub_orm
 
 from remoteappmanager.db import orm
+from remoteappmanager.db.interfaces import ABCAccounting
 from remoteappmanager.handlers.api import HomeHandler
 from remoteappmanager.logging.logging_mixin import LoggingMixin
 from remoteappmanager.docker.container_manager import ContainerManager
@@ -23,7 +24,7 @@ class Application(web.Application, LoggingMixin):
 
     user = Instance(User, allow_none=True)
 
-    db = Instance(orm.Database, allow_none=True)
+    db = Instance(ABCAccounting, allow_none=True)
 
     def __init__(self,
                  command_line_config,
@@ -209,23 +210,9 @@ class Application(web.Application, LoggingMixin):
 
     def _user_init(self):
         """Initializes the user at the database level."""
-        session = self.db.create_session()
-
         user_name = self.command_line_config.user
-        user = User(name=user_name)
-        with orm.transaction(session):
-            try:
-                user.orm_user = session.query(orm.User).filter_by(
-                                    name=user_name).one_or_none()
-            except MultipleResultsFound:
-                self.log.error("Multiple results found when "
-                               "querying for username {}. This is supposedly "
-                               "impossible because the username should be a "
-                               "unique key by design.".format(
-                                   user_name))
-                raise
-
-        self.user = user
+        self.user = User(name=user_name)
+        self.user.orm_user = self.db.get_user_by_name(user_name)
 
 
 def _server_from_url(url):
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -206,7 +206,7 @@ class Application(web.Application, LoggingMixin):
 
     def _db_init(self):
         """Initializes the database connection."""
-        self.db = orm.Database(self.file_config.db_url)
+        self.db = orm.AccountingInterface(self.file_config.db_url)
 
     def _user_init(self):
         """Initializes the user at the database level."""
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -205,7 +205,7 @@ class Application(web.Application, LoggingMixin):
 
     def _db_init(self):
         """Initializes the database connection."""
-        self.db = orm.AccountingInterface(self.file_config.db_url)
+        self.db = orm.AppAccounting(self.file_config.db_url)
 
     def _user_init(self):
         """Initializes the user at the database level."""
--- a/remoteappmanager/db/orm.py
+++ b/remoteappmanager/db/orm.py
@@ -130,7 +130,7 @@ class Database(LoggingMixin):
         Base.metadata.create_all(self.engine)
 
 
-class AccountingInterface(ABCAccounting):
+class AppAccounting(ABCAccounting):
 
     def __init__(self, url, **kwargs):
         self.db = Database(url, **kwargs)
--- a/remoteappmanager/db/orm.py
+++ b/remoteappmanager/db/orm.py
@@ -8,6 +8,7 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker, relationship
 
 from remoteappmanager.logging.logging_mixin import LoggingMixin
+from remoteappmanager.db.interfaces import ABCAccounting
 
 Base = declarative_base()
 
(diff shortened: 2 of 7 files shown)


remoteappmanager/handlers/home_handler.py
```diff
@@ -244,13 +239,13 @@ def _start_container(self, orm_user, app, policy, mapping_id):
 
         Parameters
         ----------
-        orm_user : orm.User
-            the orm user object
+        orm_user : User
+            database's user object (e.g. current_user.orm_user)
```
kitchoi: @itziakos 
@stefanoborini and I discussed offline whether we still need to pass the database's User object to `HomeHandler._start_container` or whther we can simply pass the username here (or not, username is accessible from HomeHandler.current_user).  We don't yet have a conclusion.

Note that there is no ABC defined for the database's User object (IMHO there is no need for one)

stefanoborini:
Imho, the whole interface is meant to deal with high level objects. All our controller code is supposed to directly use high level model objects. Service code (such as the container manager) works in low level entities (strings and similar) and returns its own "model orm" (e.g. Container/Image objects). The container manager acts as a DBMS for the "docker database".

kitchoi:
I also think that the discussion on the signature of  `HomeHandler._start_container(user,...)` is not blocking.  Please review the other part of the PR.

remoteappmanager/application.py
```diff
@@ -23,7 +23,7 @@ class Application(web.Application, LoggingMixin):
 
     user = Instance(User, allow_none=True)
 
-    db = Instance(orm.Database, allow_none=True)
+    db = Instance(ABCAccounting, allow_none=True)
```
stefanoborini: Ok, at this point, I think that the `db` name is deceiving. Also, what happens when we decide to handle more tasks than just accounting in the database?

remoteappmanager/application.py
```diff
@@ -205,27 +205,13 @@ def _reverse_proxy_init(self):
 
     def _db_init(self):
         """Initializes the database connection."""
-        self.db = orm.Database(self.file_config.db_url)
+        self.db = orm.AccountingInterface(self.file_config.db_url)
```
stefanoborini: I wouldn't go with Interface. Interface has a clear connotation of a fully abstract class. If it's a facade, gateway, manager, call it so. _namingthingsishard_ maybe AccountingDb? AccountingManager? Accountant?

remoteappmanager/application.py
kitchoi: I agree that the name isn't great.  `accounting`?

> what happens when we decide to handle more tasks than just accounting in the database?

It seems that the implementation does not stop you from adding more methods to ABCAccounting.  The question is just whether the name `ABCAccounting` would still be appropriate for those future tasks I don't know what.  Is this only a naming issue then?  Or there is more to it?

remoteappmanager/db/orm.py
```diff
@@ -129,6 +130,54 @@ def reset(self):
...
+
+        # We keep the same session for all transactions
+        # so that the session remains on one thread
+        self.session = self.db.create_session()
+
+    def get_user_by_name(self, user_name):
+        """ Return an orm.User given a user name.  Return None
+        if the user name is not found in the database
+        """
+        with transaction(self.session):
```
stefanoborini: Not now, but we will have to consider subtransactions sooner or later... This will isolate the transaction in a different session, with potential problems if another session/transaction is assuming some things.

Note that being readonly does not protect us. We can still have race conditions because we lose isolation if we terminate the transaction and resume it later. 

This is a general problem in our design though, so I am not too concerned at the moment, but it needs a bit more care in the future iterations.

remoteappmanager/db/orm.py
```diff
@@ -129,6 +130,54 @@ def reset(self):
...
 
 
+class AccountingInterface(ABCAccounting):
+
+    def __init__(self, url, **kwargs):
+        self.db = Database(url, **kwargs)
+
+        # We keep the same session for all transactions
+        # so that the session remains on one thread
+        self.session = self.db.create_session()
```
stefanoborini: The idea is that the session is handed outside, and every operation is within the context of a session. This way, you assume there's only one context, and you never allow it outside.

remoteappmanager/db/orm.py
```diff
@@ -129,6 +130,54 @@ def reset(self):
...
+    def get_user_by_name(self, user_name):
+        """ Return an orm.User given a user name.  Return None
+        if the user name is not found in the database
+        """
+        with transaction(self.session):
+            return self.session.query(User).filter_by(
+                name=user_name).one_or_none()
+
+    def get_apps_for_user(self, user):
+        """ Return a tuple of tuples, each containing an application
```
stefanoborini: Why not a list of tuples?

remoteappmanager/application.py
kitchoi: I have no clue, *namingthingsishardindeed*

remoteappmanager/application.py
stefanoborini: `Database` ?

remoteappmanager/db/orm.py
kitchoi: The results of this method is a state, and should be immutable.  Performance wise, tuples reuse previously used memory block of the same size.

stefanoborini: Fair enough, but I tend to interpret a tuple as something that would be a row in a db table and you tend to unpack, and a list as something that is along the columns of a database table and you tend to iterate upon.

remoteappmanager/application.py
stefanoborini: Database?

kitchoi: `orm.Database` already exists (so does `orm.Accounting`)  All the good names are taken.

stefanoborini: then change the namespace...
orm.Database is clearly a sqlalchemy oriented database, and implements the Database interface with an implementation that is sqlalchemy oriented.
You will have another implementation that handles csv files. so you need a db namespace, and two db backends: sqlalchemy and csv.

stefanoborini: Don't worry about moving things around. We need to evolve a proper schema to organise the code.

remoteappmanager/db/orm.py
kitchoi: I had to do this otherwise when calling `get_apps_for_user` from any coroutine of the HomeHandler, sqlalchemy complains about accessing the session from a different thread.

kitchoi: Scopes of a session and a transaction do not need to be the same.

tests/db/test_orm.py
```diff
@@ -2,7 +2,9 @@
 import unittest
 
 from remoteappmanager.db import orm
-from remoteappmanager.db.orm import Database, transaction, Accounting
+from remoteappmanager.db.orm import (Database, transaction, Accounting,
+                                     AccountingInterface)
```
itziakos: (suggestion): I tend to not name concrete implementations as `Interface`. how about `UserAccounting`

itziakos: We might consider naming the ABCAccounting to `User` or `AppAccounting`. We should revisit naming on Friday

remoteappmanager/db/orm.py
stefanoborini: You are supposed to create a session and do multiple transactions with them, potentially, but in our little application, it's probably a good idea to start the session, start the transaction, complete the transaction and never reuse the session again.

so one thing that you could do, is to have a begin and end methods in your database application. In the sqlalchemy one it creates the session, in the csv one it does nothing.

itziakos: We could augment the ABCAccounting API with `setup`, and `cleanup` methods if we find that we need to manage resources

kitchoi:
I have at least renamed the orm.AccountingInterface to orm.AppAccounting
Yes let's revisit naming this friday.

remoteappmanager/application.py
kitchoi: Let's discuss this naming problem on friday.

kitchoi:
@stefanoborini Please review again.

PR #52, opened 2016-07-06
Refactor remote access

stefanoborini:
Part of release 0.4.0 but I suspended work on #48 until the database is settled to prevent conflicts.
Opening for comments.

What I am doing is extract the reverse proxy and the hub as independent classes, configured appropriately. We then use methods to perform remote requests.

Note that this required the container to know its own absolute url path. The container manager cannot know this (it does not know anything about the application) so this information must be set from the outside.


stefanoborini:
Do not merge for 0.3.0 unless we get clearance from @itziakos. Also, this PR might conflict with the changes for the CSV support, so it should be merged only after that one.

stefanoborini:
@kitchoi review pls

remoteappmanager/docker/container.py
```diff
@@ -32,12 +32,22 @@ class Container(HasTraits):
...
+    base_urlpath = Unicode(None, allow_none=True)
+
     @property
-    def url(self):
+    def urlpath(self):
         """Returns the relative url of the Container."""
         return "containers/{}".format(self.url_id)
 
     @property
+    def absurlpath(self):
```
kitchoi: the `absurlpath` has quite an absurd name, maybe `abs_urlpath` is better?

remoteappmanager/docker/container.py
```diff
@@ -32,12 +32,22 @@ class Container(HasTraits):
     #: the id that will go in the URL of the container
     url_id = Unicode()
 
+    #: The absolute base prefix path for the container.
+    base_urlpath = Unicode(None, allow_none=True)
```
kitchoi: If `absurlpath` raises an Error when base_urlpath is None, why do we allow_none here?

stefanoborini: because when it's created, it is None. The creation happens at the container manager level, but the container manager has no clue of where the application is based on. This is really something I don't like, but I could not find a decent solution for it. I don't want the container manager to know the application, nor the application base. One alternative would be to tell the reverse proxy to add the base url, but then you need the absolute url regardless, when you have to do the redirection.

stefanoborini: I honestly don't have a clear answer to this problem. I see that a container should know its absolute path, but it is unable to do so when created. It's information that needs to be added later, because it's simply not available until later.

remoteappmanager/application.py
```diff
@@ -67,6 +67,65 @@ def command_line_config(self):
...
+        """Initializes the Hub instance."""
+        return Hub(endpoint_url=self.command_line_config.hub_api_url,
+                   api_key=os.environ.get('JPY_API_TOKEN', "")
+                   )
+
+    @default("db")
+    def _db_default(self):
+        """Initializes the database connection."""
+        if self.file_config.db_url.endswith('.db'):
+            self.db = orm.AppAccounting(self.file_config.db_url)
```
kitchoi: `return` instead of `self.db`

remoteappmanager/docker/container.py
stefanoborini: I agree and I loathe it as well, but I tried to stay compatible with os.path.abspath.

stefanoborini: Actually, I think that we should use urlpath() for the absolute path and that's it, but again, I am not sure of the direction of this interface.

kitchoi: For the name of `absurlpath`, if we can cope with `base_urlpath`, that would seem that another `abs_urlpath` is okay.

As for where we should compute `urlpath`, I would think a `get_container_url` as a BaseHandler method maybe good place for it.  After all it is web related.

stefanoborini: >  I would think a get_container_url as a BaseHandler method maybe good place for it.

You can't. The ReverseProxy class has an add_container() method. you pass the container, so the container must know its full absolute path. Unless we make the ReverseProxy aware of the application, which also sounds weird, or we pass the base path to the add_container method, or, as I said, pass a base path to the ReverseProxy class so that it gets appended to every request (and the container only knows its relative path).

I don' t have a strong opinion on any of these options. They are all weird in some way or another.

remoteappmanager/handlers/home_handler.py
```diff
@@ -326,7 +328,7 @@ def _wait_for_container_ready(self, container):
         server_url = "http://{}:{}{}/".format(
             container.ip,
             container.port,
-            self.application.container_url_abspath(container))
+            container.absurlpath)
```
kitchoi: here too...:(

stefanoborini: mh... right... without home handler tests we can't see this.

kitchoi:
Looks good to me.  Can merge when CI tests are green.

Issue #41, opened 2016-07-04
Support defining user, app and policy restrictions using a csv file


itziakos:
We should support alternative ways for defining the necessary user, app and policy information. For example the necessary information can be defined in a csv file. This setup would be useful for testing.

kitchoi:
Close via #61

PR #61, opened 2016-07-07
Feature csv database interface

kitchoi:
Provides for #41

### DIFF
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -7,7 +7,7 @@ import tornado.ioloop
 from jinja2 import Environment, FileSystemLoader
 from jupyterhub import orm as jupyterhub_orm
 
-from remoteappmanager.db import orm
+from remoteappmanager.db import csv_db, orm
 from remoteappmanager.db.interfaces import ABCAccounting
 from remoteappmanager.handlers.api import HomeHandler
 from remoteappmanager.logging.logging_mixin import LoggingMixin
@@ -205,7 +205,13 @@ class Application(web.Application, LoggingMixin):
 
     def _db_init(self):
         """Initializes the database connection."""
-        self.db = orm.AppAccounting(self.file_config.db_url)
+        if self.file_config.db_url.endswith('.db'):
+            self.db = orm.AppAccounting(self.file_config.db_url)
+        elif self.file_config.db_url.endswith('.csv'):
+            self.db = csv_db.CSVAccounting(self.file_config.db_url)
+        else:
+            raise ValueError("Unsupported database format: {}".format(
+                self.file_config.db_url))
 
     def _user_init(self):
         """Initializes the user at the database level."""
--- a/remoteappmanager/db/csv_db.py
+++ b/remoteappmanager/db/csv_db.py
@@ -64,12 +64,6 @@ class CSVUser(object):
     def __init__(self, name):
         self.name = name
 
-    def __eq__(self, other):
-        return self.name == other.name
-
-    def __hash__(self):
-        return hash(self.name)
-
 
 _HEADERS = ('user.name',
             'application.image',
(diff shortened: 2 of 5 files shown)


remoteappmanager/application.py
```diff
@@ -205,7 +205,13 @@ def _reverse_proxy_init(self):
 
     def _db_init(self):
         """Initializes the database connection."""
-        self.db = orm.AppAccounting(self.file_config.db_url)
+        if self.file_config.db_url.endswith('.db'):
+            self.db = orm.AppAccounting(self.file_config.db_url)
+        elif self.file_config.db_url.endswith('.csv'):
+            self.db = csv_db.CSVAccounting(self.file_config.db_url)
```
stefanoborini: Know that it's temporary, but I would check if it's a sqlalchemy path vs filesystem path first.

kitchoi: CSVAccounting would raise an IOError anyway and that would be subsequently logged.  Or do you mean it should something else in addition to raising an IOError when the path is not pointing to a valid file on the filesystem?

kitchoi: correction, it is the 'FileNotFoundError'

kitchoi:
@stefanoborini Please review

remoteappmanager/db/csv_db.py
```diff
@@ -0,0 +1,149 @@
...
+# FIXME: The user object could simply be a str (the username)
+# We need this because HomeHandler._start_container takes an object with
+# the username in its `name` attribute
+class CSVUser(object):
+
+    def __init__(self, name):
+        self.name = name
+
+
+_HEADERS = ('user.name',
```
stefanoborini: Just add a comment that it's the headers in the csv file. Trivial, but worth clarifying.

remoteappmanager/db/csv_db.py
```diff
@@ -0,0 +1,149 @@
...
+
+        Parameters
+        ----------
+        csv_file_path : str
+            File path for the CSV file
+
+        Optional parameters
+        -------------------
+        kwargs : dict
+            optional parameters for open(csv_file_path, **kwargs)
```
stefanoborini: Don't put ** in the docstrings. sphinx gets confused and thinks it's the beginning of a strong section, can't find the end, and complains in an undefined place later on.

tests/db/test_csv_db.py
```diff
@@ -0,0 +1,132 @@
...
+import unittest
+
+from remoteappmanager.db.csv_db import (
+    CSVApplication, CSVApplicationPolicy, CSVUser, CSVAccounting)
+
+from tests.db.abc_test_interfaces import ABCTestDatabaseInterface
+from tests.temp_mixin import TempMixin
+
+
+class GoodTable:
```
stefanoborini: I wonder if it would be better just to have three files in the fixtures directory instead.

remoteappmanager/db/csv_db.py
kitchoi: Good point.  Should have put `**kwargs : dict` instead, I think sphinx can handle that, no?

stefanoborini: nope... It was source of troubles. that, and colons (e.g. a comment containing ro: read only, was triggering a warning that a strong section start was found without and end, despite strong is started with ** and there were no ** in sight).

kitchoi: Apparently you could: http://www.sphinx-doc.org/en/stable/ext/example_numpy.html
But I also remember running into similar problem related to `**`

stefanoborini: So for safety, steer clear of *, ** and : (with the exception of the #: marker)

remoteappmanager/db/csv_db.py
```diff
@@ -30,7 +30,7 @@
     If undefined, common data volume is not available
 
 policy.volume_mode : str
-   Mode for read/write access (ro: read-only, rw: read-write)
```
stefanoborini: This wasn't needed, I already fixed it in the other branch. We'll get a conflict but we'll fix it. Leave it as is.

kitchoi:
@stefanoborini Since you mention the doc, I am checking against sphinx at the moment... pushing a few minor fixes soon

kitchoi:
@stefanoborini Updated sphinx doc.  And what should I do with https://github.com/simphony/simphony-remote/pull/61#discussion_r69883421?

PR #76, opened 2016-07-11
Makefile for basic deployment operations

### LINKED ISSUE #68: Provide a makefile for simple deployment operations
I would like to be able to do:

make apt-dependencies
make simphony-remote
make check

stefanoborini:
Fixes #68

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,34 +17,32 @@ deps:
 	sudo apt-get update
 	sudo apt-get install docker-engine npm nodejs-legacy python3-pip python3.4-venv
 	sudo npm install -g configurable-http-proxy
+	python3 -mvenv venv
+	. venv/bin/activate && pip3 install docker-py
+
+.PHONY: develop
+develop: 
+	@echo "Installing application"
+	@echo "----------------------"
+	. venv/bin/activate && python3 setup.py develop 
 
 .PHONY: install
 install: 
 	@echo "Installing application"
 	@echo "----------------------"
-	python3 -mvenv venv
-	. venv/bin/activate \
-		&& pip3 install docker-py \
-		&& python3 setup.py develop 
+	. venv/bin/activate && python3 setup.py install
 
 .PHONY: certs
-certs: jupyterhub/test.crt
-
-jupyterhub/test.crt:
+certs: 
 	@echo "Creating certificates"
 	@echo "---------------------"
-	pushd scripts \
-		&& sh generate_certificate.sh \
-		&& cp test.* ../jupyterhub/ \
-		&& popd
+	-pushd jupyterhub && sh ../scripts/generate_certificate.sh && popd
 
 .PHONY: db
-db: $(HOME)/remoteappmanager.db
-
-$(HOME)/remoteappmanager.db:
+db: 
 	@echo "Creating database"
 	@echo "-----------------"
-	. venv/bin/activate && remoteappdb --db=~/remoteappmanager.db init
+	-. venv/bin/activate && remoteappdb --db=~/remoteappmanager.db init
 
 .PHONY: images
 images:
@@ -58,4 +56,4 @@ images:
 test:
 	@echo "Running testsuite"
 	@echo "-----------------"
-	. venv/bin/activate && python -m tornado.testing discover
+	-. venv/bin/activate && python -m tornado.testing discover
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,53 @@
+SHELL=bash
 OS=$(shell uname)
 
 .PHONY: help
 help:
 	@echo "Please specify one of the following:"
 	@echo ""
+	@echo "all: full deployment"
 	@echo "deps: install the dependencies of the project for your platform"
 	@echo "install: install the application"
+	@echo "certs: builds and installs the certificates"
+	@echo "db: initializes the sqlite database"
 	@echo "check: performs sanity checks for the installation"
+	@echo ""
 
 .PHONY: deps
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+.PHONY: db
+db: $(HOME)/remoteappmanager.db
+
+$(HOME)/remoteappmanager.db:
+	@echo "Creating database"
+	@echo "-----------------"
+	. venv/bin/activate && remoteappdb --db=~/remoteappmanager.db init
+
+.PHONY: images
```
kitchoi: do we need a make target for the images?  seems to me trivial that not worth maintaining.

Makefile
```diff
@@ -0,0 +1,55 @@
...
+install: 
+	@echo "Installing application"
+	@echo "----------------------"
+	python3 -mvenv venv
+	. venv/bin/activate \
+		&& pip3 install docker-py \
+		&& python3 setup.py develop 
+
+.PHONY: certs
+certs: jupyterhub/test.crt
```
kitchoi: Why the redirection of targets here? (likewise for `db`)

Makefile
```diff
@@ -0,0 +1,55 @@
...
+	sudo npm install -g configurable-http-proxy
+
+.PHONY: install
+install: 
+	@echo "Installing application"
+	@echo "----------------------"
+	python3 -mvenv venv
+	. venv/bin/activate \
+		&& pip3 install docker-py \
+		&& python3 setup.py develop
```
kitchoi: shall we do `pip install git+https://...@{version}` instead where `version` is either set by an environment variable or a default?  This way one can pass around the make file without having git clone the repository first.

Makefile
stefanoborini: I don't want to recreate the certificates if they are already there.

Makefile
stefanoborini: So to install everything from the makefile? I am not sure it's what we want. or do we?
If so, then we need a develop: action to simply build the local setup and insert it as develop package.

kitchoi: why develop?

kitchoi:
According to #68, there should be a target for testing as well.

Makefile
stefanoborini: mirrors setup.py develop

Makefile
stefanoborini: It doesn't hurt, especially if they want to get started soon.

As part of this PR, I should probably also modify the deployment documentation. When the current rules are ok, I'll move on to fix that as well.

Makefile
kitchoi: I mean why do we want the `install` target to do `setup.py develop`.  Should we have a develop target for this instead?

Makefile
kitchoi: That seems to me should be dealt with by `generate_cerficiate.sh`

kitchoi: If you agree, we can also fix this later ;)

Makefile
```diff
@@ -0,0 +1,61 @@
...
+.PHONY: certs
+certs: jupyterhub/test.crt
+
+jupyterhub/test.crt:
+	@echo "Creating certificates"
+	@echo "---------------------"
+	pushd scripts \
+		&& sh generate_certificate.sh \
+		&& cp test.* ../jupyterhub/ \
+		&& popd
```
kitchoi: Minor, subjective comment, one could do 
`pushd jupyterhub/ && sh ../scripts/generate_cerficate.sh && popd`
The slight difference is that there is no test.* file left behind in ./scripts, and `generate_cerificate.sh` can check whether the cert files already exist.  The current command will do too.

Makefile
kitchoi: Instead of doing a `pip3 install` for the `install` target, I'd prefer the make target be called `develop` for `setup.py develop`.  Maybe that's easier for now.

Makefile
```diff
@@ -0,0 +1,61 @@
...
+	python3 -mvenv venv
+	. venv/bin/activate \
+		&& pip3 install docker-py \
+		&& python3 setup.py develop 
+
+.PHONY: certs
+certs: jupyterhub/test.crt
+
+jupyterhub/test.crt:
+	@echo "Creating certificates"
```
kitchoi: "Creating certificates in ./jupyterhub/"

Makefile
stefanoborini: Not really, generate_certificate.sh creates a certificate in the local directory, whatever that is, and dumps a new certificate in there.
this make rule not only creates the certificate, it also "installs" the certificate for the current deployment. It's the "install" part that I want to prevent to happen twice (and as a consequence, the execution of the generate_certificate script).

kitchoi: Please see https://github.com/simphony/simphony-remote/pull/76/files/8d5b4f40c826cf29599385789da7eb29fa1753a8#r70289426

kitchoi:
Minor suggestions
1. Rename `install` target to `develop` to reflect the fact that we are doing a `setup.py develop`
2. Clean up certificates created in ./scripts
3. Make clear where the certificates are eventually placed (./jupyterhub)

kitchoi:
> Minor suggestions
1. Rename install target to develop to reflect the fact that we are doing a setup.py develop
2. Clean up certificates created in ./scripts
3. Make clear where the certificates are eventually placed (./jupyterhub)

The rest is good to merge. 
Thanks @stefanoborini

kitchoi:
Thanks @stefanoborini, please feel free to merge when green.

Issue #62, opened 2016-07-07
verify_token should return True or False, not raise HTTPError


stefanoborini:
The signature of Hub.verify_token is a bit inconsistent. I suspect we use it for returning a reason for the failure, but it just looks weird, it might need some thought.

kitchoi:
In jupyterhub, it is okay to raise HTTPError because their BaseHandler has overwritten the `write_error` and render an error page with the HTTPError messages.  We could do the same with `write_error`.
As for successful request, it should return the json data instead.

PR #77, opened 2016-07-11
verify_token always return a dict and not raise an HTTPError

kitchoi:
`verify_token` would return a dict containing the user data from jupyterhub and the remoteappmanager would further verify the data.  The dictionary is empty if authentication/authorisation fails.

HTTPError is not raised, but in the future we may want to introduce the HTTPError back again for better error message on the user's end.

### DIFF
--- a/remoteappmanager/handlers/base_handler.py
+++ b/remoteappmanager/handlers/base_handler.py
@@ -3,7 +3,6 @@ from urllib.parse import urljoin
 from tornado import web
 
 from remoteappmanager.logging.logging_mixin import LoggingMixin
-from tornado.httpclient import HTTPError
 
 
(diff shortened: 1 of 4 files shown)


kitchoi:
Fix #62

stefanoborini:
I need this one merged for the async http request.

stefanoborini:
@kitchoi still broken.

kitchoi:
@stefanoborini just pushed a fix as you commented :)

stefanoborini:
@kitchoi good for me. please merge when green. I'll take care of the conflicts with the other branch (I still have some work to do there)

kitchoi:
Thanks @stefanoborini

PR #45, opened 2016-07-04
Fix for lack of url_id in viewing operation

stefanoborini:
Adds a label to the container with the url_id, so that we can recover it.

PR #79, opened 2016-07-11
Fix loading docker image labels

### LINKED ISSUE #64: Image's ui_name are no longer shown
Docker images with a `ui_name` label no longer has its name shown on the home page of the singleuser app.
It is a retrogression.

kitchoi:
Fix #64

### DIFF
--- a/remoteappmanager/docker/image.py
+++ b/remoteappmanager/docker/image.py
@@ -24,7 +24,14 @@ class Image(HasTraits):
     @classmethod
     def from_docker_dict(cls, docker_dict):
         """Converts the dict response from the dockerpy library into an
-        instance of this class, extracting the relevant information."""
+        instance of this class, extracting the relevant information.
+
+        Parameters
+        ----------
+        docker_dict : dict
+           Results of `docker.client.inspect_image` or an item of the
+           result of `docker.client.images`
+        """
         self = cls()
         self.docker_id = docker_dict["Id"]
 
--- a/remoteappmanager/docker/image.py
+++ b/remoteappmanager/docker/image.py
@@ -33,7 +33,9 @@ class Image(HasTraits):
         except (KeyError, IndexError):
             self.name = ''
 
-        labels = docker_dict.get("Labels")
+        labels = (docker_dict.get("Labels") or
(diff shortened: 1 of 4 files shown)


remoteappmanager/docker/image.py
```diff
@@ -33,7 +33,9 @@ def from_docker_dict(cls, docker_dict):
         except (KeyError, IndexError):
             self.name = ''
 
-        labels = docker_dict.get("Labels")
+        labels = (docker_dict.get("Labels") or
+                  docker_dict.get("Config", {}).get("Labels"))
+
```
stefanoborini: If I remember correctly, we decided to use a consistent strategy for this, that is, not to pass the images() dict, but the inspect_image dict instead. We already encountered this problem with the containers IIRC. Please check if we are consistent.

kitchoi: > If I remember correctly, we decided to use a consistent strategy for this, that is, not to pass the images() dict, but the inspect_image dict instead.

I am not aware of this decision being made collectively.  Can you point me to the relevant issue please?

stefanoborini:
@kitchoi https://github.com/simphony/simphony-remote/pull/45/files

kitchoi:
@sbo With the current change, `from_docker_dict` can take dictionary from both `images` and `inspect_image`.  I can make it such that it is only compatible with `inspect_image`.  IMHO, I like the dual compatibility (also realise I should add a test for this).  What do you think?

stefanoborini:
@kitchoi I don't have a strong opinion on it, but I do see a problem from the code client side, that you don't know what you are passing. I prefer methods to convey exactly what they operate on, but that's just me. So, your choice, but whatever the choice is, we need to do the same for the container, otherwise the API will differ between the two. Not in this PR, open an issue for it and we'll fix in the slack at the end of the week (if any)

kitchoi:
@sbo I will open an issue for consistency between `docker.container.Container.from_docker_dict` and `docker.image.Image.from_docker_dict`.  I also just added a description in the docstring for the latter.

kitchoi:
@sbo good to merge?

PR #80, opened 2016-07-11
Rest api infrastructure [WIP]

stefanoborini:
Basic infrastructure for the REST API. Will not provide the actual API, just the building blocks to build the actual API.

Still needed is the logic that interacts with docker, extracts the current images, decides for a representation, etc..

Opening for comments.


stefanoborini:
Note: coverage is only apparently going down, because we added code that is tested, but through an HTTP connection. As such, it's not detected by coverage because it's running as a separate process via AsyncHTTPTestCase.

kitchoi:
I agree with #87 that the REST API can be an external package and the remoteappmanager is used as a library.  That way only one rest service is run for all singleuser-app processes, and we are not necessarily restricted to using tornado.  For example one can use http://python-eve.org/

remoteappmanager/restmodel/image.py
```diff
@@ -0,0 +1,9 @@
+from remoteappmanager.rest.resource import Resource
+from tornado import gen
+
+
+class Image(Resource):
+    @classmethod
+    @gen.coroutine
+    def list(self):
```
kitchoi: You meant `items` instead of `list`?

stefanoborini:
We are not limited to tornado, but we would have to integrate both tornado and flask in our setup, and configure a nginx in front that dispatches to the appropriate application. I am not sure we want to do that.

eve is a much more complicated piece of code. It's very flexible, but I seriously doubt we need that amount of flexibility.

remoteappmanager/rest/registry.py
```diff
@@ -0,0 +1,52 @@
+from remoteappmanager.rest.resource import Resource
+
+
+class Registry:
+    def __init__(self):
+        self._registered_types = {}
+
+    def register(self, typ):
+        """Registers a Resource type with an appropriate
```
kitchoi: Why use an under variable (__collection_name__) instead of passing collection name as an argument in this method?

stefanoborini: similar to sqlalchemy ``__table__``

kitchoi: But `register` has to be specifically called anyway... It is unlikely future data model would have `__collection_name__` so that's no big deal, just wondering.

kitchoi:
I suppose any uncaught error raises HTTP Error 500?

remoteappmanager/rest/resource.py
```diff
@@ -0,0 +1,166 @@
...
+        ----------
+        identifier: str
+            A string identifying the resource
+
+        Returns
+        -------
+        bool: True if found, False otherwise.
+        """
+        try:
+            cls.retrieve(identifier)
```
kitchoi: yield

stefanoborini:
Unhandled exceptions raise 500 as they should. They are technically internal server errors, and tornado raises a 500 by default.

remoteappmanager/rest/resource.py
```diff
@@ -0,0 +1,166 @@
...
+    and reimplement the CRUD class methods with the appropriate
+    logic.
+    """
+    @classmethod
+    @gen.coroutine
+    def create(cls, representation):
+        """Called to create a resource with a given representation
+        The representation is a dictionary containing keys. The
+        reimplementing code is responsible for checking the validity
+        of the representation.
```
kitchoi: Called by POST method of a collection of the resource.

kitchoi:
Would handling of authentication tokens cause a fundamental change of the framework?  For example, GET `/api/v1/images` may require authentication whilst GET `/api/v1/containers` may not, or they may authenticate different things.  `tornado.web.authenticated` can only decorate RequestHandler's method, right?

kitchoi:
> Would handling of authentication tokens cause a fundamental change of the framework? For example, GET /api/v1/images may require authentication whilst GET /api/v1/containers may not, or they may authenticate different things. tornado.web.authenticated can only decorate RequestHandler's method, right?

Tokens would need to be placed in the request body, which would need to be passed to the Resource's methods.

stefanoborini:
@kitchoi web.authenticated can indeed only decorate handler methods. ACL should probably be on the Resource object. The handler issues a Resource.can_perform(action, user) and act accordingly, or we should have the individual methods accept the user as well as part of their signature.

kitchoi:
Would it be simplier if the request body be passed to the Resource at all time?

remoteappmanager/rest/resource.py
```diff
@@ -0,0 +1,166 @@
...
+        NotImplementedError:
+            If the resource does not support the method.
+        """
+        raise NotImplementedError()
+
+    @classmethod
+    @gen.coroutine
+    def retrieve(cls, identifier):
+        """Called to retrieve a specific resource given its
+        identifier.
```
kitchoi: Called by GET method of the resource's uri

remoteappmanager/rest/resource.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Parameters
+        ----------
+        identifier: str
+            A string identifying the resource
+
+        Returns
+        -------
+        representation: dict
+            a dict representation of the resource.
```
kitchoi: The representation is written directly (without modification) to the response body.

stefanoborini:
@kitchoi you would violate one assumption that I am trying to preserve: isolation of HTTP concern from Resource concern. Passing the resource object would violate that.

remoteappmanager/rest/resource.py
stefanoborini: tornado renders dicts and lists automatically as json.

kitchoi: Good. Thanks.

stefanoborini:
@kitchoi On top of this PR, I think I'll try to implement a basic API to retrieve the currently available images. Let's see how it behaves with the most simple case. Authentication aside, it should be straightforward.

kitchoi:
We are already passing request body renamed as `representation`, it seems the same if we also pass the request header (correction to my previous comment, authorization data is passed in header).  Renaming `representation` back to`request_body` would also be implicitly transparent (!?)

stefanoborini:
@kitchoi the request body is not the representation. The request body can be XML, and it's the HTTP rendering of the representation. The internal representation is a simple dict.

kitchoi:
@sbo that's what is written down in the doc, but in practice: https://github.com/simphony/simphony-remote/pull/80/files#diff-a14a05e805bd33fd47fad4cd421fa18cR121 (another instance in `create`)

stefanoborini:
create receives an internal representation. That part you highlight converts the HTTP representation to internal representation (by decoding the json document)

stefanoborini:
@kitchoi  Ok, I think I will do the following changes:

Resource will have instance methods, and will be instantiated. The application will be passed to the resources at initialization, so that in the CRUD handlers we have access to the application, and therefore to everything else. This is the same pattern that tornado.web.RequestHandler uses. 

The alternative, setting the application as a class member, is awkward.

stefanoborini:
Closing this one. The reference implementation is now #90, which is built on top of this one.

PR #82, opened 2016-07-11
Asynchronous verification of the user.

### LINKED ISSUE #37: Async the hub call for user verification
Currently, the hub is inquired synchronously for the user, meaning that we might get lag if the hub is unreachable. It's unlikely as a scenario, but we prefer to have everything asynchronous.

stefanoborini:
Fixes #37

### DIFF
--- a/tests/handlers/test_home_handler.py
+++ b/tests/handlers/test_home_handler.py
@@ -1,7 +1,6 @@
 import os
 import urllib.parse
 from unittest import mock
-import socket
 
 import tornado.netutil
 import tornado.testing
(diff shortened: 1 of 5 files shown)


tests/utils.py
```diff
@@ -275,6 +279,35 @@ def invocation_argv():
...
+                                        '127.0.0.1',
+                                        family=socket.AF_INET,
+                                        reuse_port=reuse_port)[0]
+    port = sock.getsockname()[1]
+    return sock, port
+
+
+class AsyncHTTPTestCase(tornado.testing.AsyncHTTPTestCase):
+    """Base class workaround for the above condition."""
+    def setUp(self):
```
kitchoi: Maybe we don't want to overload `setUp` here given we frequently subclass `TempMixin` and use the `TempMixin.setUp` instead.  How about making this a module-level function that takes a test case as an argument?

kitchoi:
Except for one comment above, the rest looks good to me.

tests/utils.py
stefanoborini: I don't understand. the setUp will be called according to the mro...

kitchoi: Ah right, `TempMixin` calls `super().setUp()` too.  But I saw you did the same `bind_unused_port` handling in `TestHomeHandler.setUp`, so you have set it twice?

tests/handlers/test_home_handler.py
```diff
@@ -46,13 +29,13 @@ def coro(*args, **kwargs):
...
 
-class TestHomeHandler(TempMixin, AsyncHTTPTestCase):
+class TestHomeHandler(TempMixin, utils.AsyncHTTPTestCase):
     def setUp(self):
         self._old_proxy_api_token = os.environ.get("PROXY_API_TOKEN", None)
         os.environ["PROXY_API_TOKEN"] = "dummy_token"
 
         self._bind_unused_port_orig = tornado.testing.bind_unused_port
-        tornado.testing.bind_unused_port = _bind_unused_port
+        tornado.testing.bind_unused_port = utils.bind_unused_port
```
kitchoi: utils.AsyncHTTPTestCase.setUp has this setup already so this is not needed (and line 46 as well)

Issue #71, opened 2016-07-08
Isolate the session/transaction generation code in the BaseHandler


stefanoborini:
We would like to have every transaction with the user as isolated, and to do this, we need to start the session and the transaction as soon as the handler receives the request, and terminate it when the web page is rendered.

We need to use prepare() and on_finish for this, but we have to remember that our databases can be different sources of data, some of which may not have a transaction (e.g. CSV), so that needs to be kept into account (e.g. a start(), stop() method)?

PR #89, opened 2016-07-12
Rename orm_user to account and document as opaque-type

### LINKED ISSUE #67: User.orm_user should be renamed as User.account with an opaque type
The `orm_user` attribute of the singleuser app is an opaque object handled by the database's accounting.
It should be renamed as `account` and be given an opaque type.

kitchoi:
Fix #67

### DIFF
--- a/remoteappmanager/db/interfaces.py
+++ b/remoteappmanager/db/interfaces.py
@@ -55,7 +55,7 @@ class ABCApplicationPolicy(metaclass=ABCMeta):
 
 
 class ABCAccounting(metaclass=ABCMeta):
-    """ Main accounting interface required by the single User application.
+    """ Main accounting interface required by the single user application.
     """
 
     @abstractmethod
@@ -69,16 +69,17 @@ class ABCAccounting(metaclass=ABCMeta):
 
         Returns
         -------
-        a User-like object that the Database understands
+        account : opaque-type
+           an object that the database understands
         """
 
     @abstractmethod
-    def get_apps_for_user(self, user):
-        """ Return an iterable of ApplicationConfig for a given User
+    def get_apps_for_user(self, account):
+        """ Return an iterable of ApplicationConfig for a given account
 
         Parameters
         ----------
-        user : User-like
+        account : opaque-type
            Same type as the result of `get_user_by_name`
 
         Returns
--- a/remoteappmanager/handlers/home_handler.py
+++ b/remoteappmanager/handlers/home_handler.py
@@ -78,7 +78,7 @@ class HomeHandler(BaseHandler):
         mapping_id = options["mapping_id"][0]
 
         all_apps = self.application.db.get_apps_for_user(
-            self.current_user.orm_user)
+            self.current_user.account)
 
         choice = [(m_id, app, policy)
                   for m_id, app, policy in all_apps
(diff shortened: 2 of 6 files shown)


remoteappmanager/db/interfaces.py
```diff
@@ -69,16 +69,17 @@ def get_user_by_name(self, user_name):
...
         -------
-        a User-like object that the Database understands
+        account : opaque-type
+           an object that the database understands
         """
 
     @abstractmethod
-    def get_apps_for_user(self, user):
-        """ Return an iterable of ApplicationConfig for a given User
+    def get_apps_for_user(self, account):
```
stefanoborini: At this point I wonder if we should use get_apps_for_account instead.

remoteappmanager/handlers/home_handler.py
```diff
@@ -90,9 +90,9 @@ def _actionhandler_start(self, options):
         _, app, policy = choice[0]
 
         container = None
-        orm_user = self.current_user.orm_user
+        user_name = self.current_user.name
         try:
-            container = yield self._start_container(orm_user,
+            container = yield self._start_container(user_name,
```
stefanoborini: Why did you change this?

remoteappmanager/handlers/home_handler.py
```diff
@@ -254,7 +252,6 @@ def _start_container(self, orm_user, app, policy, mapping_id):
         Container
         """
 
-        user_name = orm_user.name
```
stefanoborini: The fact that we use only one part of an object does not imply that you should just pass that part. The API is made so that it homogeneously accepts objects. Now it's no longer so.

remoteappmanager/db/interfaces.py
kitchoi: In my previous branch I did that, but think it might be too trivial to worth breaking the API that is in a release.

remoteappmanager/handlers/home_handler.py
kitchoi: what the orm_user was is now an opaque type that we do not implement an interface for, so there is no guarantee and no public api for how this opaque type should behave, i.e. we can't make sure the object has an attribute `name`.  Afterall this is a private, "helper method that works with low level data".  Is the homogeneity of types of arguments to this single method worth a ABCUser?  I don't think so.

remoteappmanager/handlers/home_handler.py
kitchoi: We have agreed that the singleuserapp's user holds an opaque type `account`, we can no longer make assumption on what it contains.

remoteappmanager/handlers/home_handler.py
stefanoborini: Ok, what's the point of an opaque object if you can't basically inquire anything about its state? How do we plan to use it?

kitchoi: The `account` object is only needed by the database and no need for anything else, its implementation is defined by the database (orm or csv_db) and the object's state is inquired there.

remoteappmanager/db/interfaces.py
stefanoborini: I don't think that anyone expects API consistency before 1.0...

kitchoi: I don't mind breaking it at all if we all agree to it.

kitchoi:
@sbo Can I merge?  I needed this for #71

PR #81, opened 2016-07-11
Add support for validating docker image name in remoteappdb

kitchoi:
Provides for #58 

Also introduced `click.testing.CliRunner` in the tests and `RemoteAppDBContext` in remoteappdb.__main__ so that we can mock and test the module (docker in this case) without relying on subprocess.
Opening an issue for applying the same strategy for other tests outside the scope of #58.

PR #94, opened 2016-07-13
Add foreign key support for sqlite

### LINKED ISSUE #56: Enable foreign key support for sqlite
Currently our orm data model, when relying on sqlite, does not enforce cascade delete or foreign keys. The problem is that by default sqlite does not enforce foreign keys. 

We need to check the docs

https://www.sqlite.org/foreignkeys.html#fk_basics

and the sqlalchemy requirements, then modify our datamodel sqlalchemy description so that cascade delete and foreign keys are enforced (e.g. when a user or application is deleted, the appropriate entries in the application policy table must be deleted as well)

kitchoi:
Fix #56

### DIFF
--- a/remoteappmanager/db/orm.py
+++ b/remoteappmanager/db/orm.py
@@ -1,8 +1,10 @@
 import contextlib
+import sqlite3
 
 from sqlalchemy import (
     Column, Integer, Boolean, Unicode, ForeignKey, create_engine, Enum,
-    literal)
+    literal, event)
+from sqlalchemy.engine import Engine
 from sqlalchemy.exc import OperationalError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker, relationship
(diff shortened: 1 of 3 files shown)


kitchoi:
WIP because its tests will break after #81 is merged.

kitchoi:
@sbo ready for review again

remoteappmanager/db/orm.py
```diff
@@ -90,6 +93,14 @@ class Accounting(Base):
     application_policy = relationship("ApplicationPolicy")
 
 
+@event.listens_for(Engine, "connect")
+def set_sqlite_pragma(dbapi_connection, connection_record):
```
stefanoborini: import sqlite3 here.

stefanoborini: I would call it _enable_foreign_keys()

remoteappmanager/db/orm.py
```diff
@@ -90,6 +93,14 @@ class Accounting(Base):
...
+    if type(dbapi_connection) is sqlite3.Connection:
+        cursor = dbapi_connection.cursor()
```
stefanoborini: with contextlib.closing(cursor):

remoteappmanager/db/orm.py
```diff
@@ -1,8 +1,10 @@
 import contextlib
+import sqlite3
```
stefanoborini: Keep it encapsulated in the enabling method, so that it works regardless if we have/use sqlite or not.

kitchoi: `sqlite3` is in the standard library...

stefanoborini: but not necessarily compiled in. You can disable it (at least it was optional in py27, I don't know if it is in py34)

remoteappmanager/db/orm.py
kitchoi: I made the function name `_set_sqlite_pragma` so that sphinx would skip it.  It is the place where future pragma can be added though so I kept the rest of the name

stefanoborini:
codecov irrelevant. merging

PR #95, opened 2016-07-13
Make sqlite database session and transaction have the same scope

### LINKED ISSUE #71: Isolate the session/transaction generation code in the BaseHandler

kitchoi:
Fix #71 

Will likely have mini merge conflicts with #94 on imports.

### DIFF
--- a/remoteappmanager/db/orm.py
+++ b/remoteappmanager/db/orm.py
@@ -3,6 +3,7 @@ import contextlib
 from sqlalchemy import (
     Column, Integer, Boolean, Unicode, ForeignKey, create_engine, Enum)
 from sqlalchemy.exc import OperationalError
+from sqlalchemy.orm.exc import DetachedInstanceError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker, relationship
 
@@ -147,7 +148,9 @@ class AppAccounting(ABCAccounting):
         # Removing internal references to the session is
         # required such that the object is detached and
         # can be reused in a different thread
-        session.expunge(user)
+        if user:
+            session.expunge(user)
+
         session.close()
         return user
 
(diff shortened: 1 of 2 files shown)


remoteappmanager/db/orm.py
```diff
@@ -214,41 +217,20 @@ def apps_for_user(session, user):
...
+    res = session.query(Accounting).join(
+        Accounting.user, aliased=True).filter_by(
+            name=user_name).all()
 
     return [(acc.application.image + "_" + str(acc.application_policy.id),
              acc.application,
              acc.application_policy) for acc in res]
-
-
-def user_can_run(session, user, application, policy):
```
kitchoi: this is not used any more even on the master.

stefanoborini:
This is not what the issue was about. What it needs to be done is to start the session at the beginning of the request, and conclude it at the end of the request, so that the whole operation is isolated. 

I don't understand what you want to achieve with this code.

kitchoi:
The goal is to "the session and the transaction as soon as the handler receives the request, and terminate it when the web page is rendered", and this is achieved by containing the session and the transaction in the same scope, i.e. in `get_user_by_name` and `get_apps_for_user`.  Note that there is no more session/transaction in HomeHandler or BaseHandler.

kitchoi:
Please also note that session and transaction are only relevant for the `db.orm` module.

kitchoi:
Please note that with this PR, the scopes of session and transaction are the tightest possible.  If #71 has meant for a bigger scope, please specify in terms of the scopes for the session and transaction for each of them, from when to when they should be kept alive, and for what purpose.  Thanks.

remoteappmanager/db/orm.py
```diff
@@ -156,18 +157,25 @@ class AppAccounting(ABCAccounting):
...
+            user = session.query(User).filter_by(
                 name=user_name).one_or_none()
 
+        # Removing internal references to the session is
+        # required such that the object is detached and
+        # can be reused in a different thread
+        if user:
+            session.expunge(user)
+
+        session.close()
```
itziakos: we need a try finally here or a context manager to make sure close is called

PR #102, opened 2016-07-14
Make sure container public port value is not None

kitchoi:
Fix bug described in #63

PR #103, opened 2016-07-14
Fix problem with OSX and View button

### LINKED ISSUE #100: View not working on mac due to incorrect IP
We found that View does not work on mac because inquiring the docker client returns a 0.0.0.0 ip. This works on linux, but on mac the container is residing on a separate virtual machine. Start works because it retrieves the information directly after starting the container, and this is handled by the container manager (which knows this) but with View, the docker client is inquired directly, so the container manager cannot do any magic.

the docker client should be considered private.

### LINKED ISSUE #101: Switch from container_id to url_id to refer to containers in the form
Currently in the form we use the container id. This is a residue that we never removed, but it's more consistent to use the url_id for all operations, and keep the container identifier hidden. 

It's trivial to fix, but we need to change the filter when inquiring the docker server to check label SIMPHONY_NS.url_id, instead of the id.

stefanoborini:
Fixes #100
Fixes #101 as a corollary

### DIFF
--- a/remoteappmanager/docker/container_manager.py
+++ b/remoteappmanager/docker/container_manager.py
@@ -165,12 +165,24 @@ class ContainerManager(LoggingMixin):
         containers = []
         infos = yield self.docker_client.containers(filters=filters)
         for info in infos:
-            container = Container.from_docker_containers_dict(info)
+            try:
+                container = Container.from_docker_containers_dict(info)
+            except Exception:
+                self.log.exception("Unable to parse container info.")
+                continue
 
             # override the ip and port obtained by the docker info with the
             # appropriate ip and port, considering that we might be using a
             # separate docker machine
-            ip, port = yield from self._get_ip_and_port(container.docker_id)
+            try:
+                ip, port = yield from self._get_ip_and_port(
+                    container.docker_id)
+            except RuntimeError:
+                self.log.exception(
+                    "Unable to retrieve ip/port "
+                    "for container {}".format(container.docker_id))
+                continue
+
             container.ip = ip
             container.port = port
             containers.append(container)
@@ -348,13 +360,25 @@ class ContainerManager(LoggingMixin):
         Return
         ------
         A tuple (ip, port)
+
+        Raises
+        ------
+        RuntimeError:
+            If for some reason it cannot retrieve the information
         """
 
         # retrieve the actual port binding
-        resp = yield self.docker_client.port(container_id, self.container_port)
+        try:
+            resp = yield self.docker_client.port(container_id,
+                                                 self.container_port)
+        except Exception as e:
+            raise RuntimeError("Failed to get port info for {}. "
+                               "Exception: {}.".format(container_id,
+                                                       str(e)))
 
         if resp is None:
-            raise RuntimeError("Failed to get port info for %s" % container_id)
+            raise RuntimeError("Failed to get port info for {}. "
+                               "Port response was None.".format(container_id))
 
         # We assume we are running on linux without any additional docker
         # machine. The container will therefore be reachable at 127.0.0.1.
@@ -367,7 +391,13 @@ class ContainerManager(LoggingMixin):
             if url.scheme == 'tcp':
                 ip = url.hostname
 
-        port = int(resp[0]['HostPort'])
+        try:
+            port = int(resp[0]['HostPort'])
+        except (KeyError, IndexError, ValueError) as e:
+            raise RuntimeError("Failed to get port info for {}. "
+                               "Exception: {}.".format(container_id,
+                                                       str(e)))
+
         return ip, port
 
     @gen.coroutine
--- a/remoteappmanager/docker/container_manager.py
+++ b/remoteappmanager/docker/container_manager.py
@@ -393,7 +393,7 @@ class ContainerManager(LoggingMixin):
 
         try:
             port = int(resp[0]['HostPort'])
-        except (KeyError, IndexError, ValueError) as e:
+        except (KeyError, IndexError, ValueError, TypeError) as e:
             raise RuntimeError("Failed to get port info for {}. "
                                "Exception: {}.".format(container_id,
                                                        str(e)))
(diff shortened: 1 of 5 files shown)


remoteappmanager/handlers/home_handler.py
```diff
@@ -142,10 +151,18 @@ def _actionhandler_view(self, options):
     def _actionhandler_stop(self, options):
         """Stops a running container.
         """
+        try:
+            url_id = options["url_id"][0]
+        except (KeyError, IndexError):
+            message = "Failed to retrieve valid url_id from form."
+            self.log.exception(message)
+            self.finish(message)
```
kitchoi: Let is raise, line 60 will handle the error, log the exception and display a user-friendly error message in the home page

remoteappmanager/handlers/home_handler.py
```diff
@@ -125,11 +124,21 @@ def _actionhandler_view(self, options):
         It is not different from pasting the appropriate URL in the
         web browser, but we validate the container id first.
         """
-        container = yield self._container_from_options(options)
+        try:
+            url_id = options["url_id"][0]
+        except (KeyError, IndexError):
+            message = "Failed to retrieve valid url_id from form."
+            self.log.exception(message)
+            self.finish(message)
```
kitchoi: Same here, let it raise.

tests/docker/test_container_manager.py
```diff
@@ -53,14 +53,34 @@ def test_containers_from_mapping_id(self):
...
+        self.manager.docker_client.client = docker_client
+
+        result = yield self.manager.container_from_url_id("url_id")
+        expected = Container(docker_id='someid',
+                             mapping_id="mapping",
+                             name='/remoteexec-image_3Alatest_user',
+                             image_name='simphony/mayavi-4.4.4:latest',  # noqa
+                             image_id='imageid',
+                             ip='127.0.0.1',
+                             port=None,
```
kitchoi: This will conflict with #102

kitchoi: I mean logically.

remoteappmanager/docker/container_manager.py
```diff
@@ -127,8 +127,55 @@ def containers_from_mapping_id(self, user_name, mapping_id):
...
+        containers = []
         infos = yield self.docker_client.containers(filters=filters)
-        return [Container.from_docker_containers_dict(info) for info in infos]
+        for info in infos:
+            container = Container.from_docker_containers_dict(info)
+
+            # override the ip and port obtained by the docker info with the
+            # appropriate ip and port, considering that we might be using a
+            # separate docker machine
+            ip, port = yield from self._get_ip_and_port(container.docker_id)
```
kitchoi: Just wondering how this particular method should behave if ip/port is not found.  The method's name and docstring do not suggest that it assumes the presence of ip and port.

kitchoi:
Some minor comments and a note on test conflict with #102 (I can fix it there if this PR if merged first), the rest looks good to me.  But a few tests fail...

remoteappmanager/docker/container_manager.py
stefanoborini: Well, it should raise probably. I'll revise the code under this assumption, and formalise one exception for it.

kitchoi: If this method `containers_from_dict` is expected to raise for nonexisting ip/port, then I think you don't need to do anything right?  `_get_ip_and_port` already raises RuntimeError

kitchoi: Sorry, `containers_from_filters`, not `containers_from_dict`

stefanoborini: Yes but I don't want the method to stop if one container raises. I want just to log that and keep processing the others.

kitchoi: Ah I see.  Make sense.

stefanoborini: Plus, the tests are broken (expected) I wanted to run it on linux and see the result.

stefanoborini:
@kitchoi we've been to liberal with linux as a development platform. It's just by accident that it worked until now. The assumption that the host is the docker machine simplified a lot of things, but in practice this is hardly a guarantee in the long term (as we would like to dispatch execution around).

remoteappmanager/docker/container_manager.py
```diff
@@ -165,12 +165,24 @@ def containers_from_filters(self, filters):
...
             # appropriate ip and port, considering that we might be using a
             # separate docker machine
-            ip, port = yield from self._get_ip_and_port(container.docker_id)
+            try:
+                ip, port = yield from self._get_ip_and_port(
+                    container.docker_id)
+            except RuntimeError:
+                self.log.exception(
+                    "Unable to retrieve ip/port "
+                    "for container {}".format(container.docker_id))
```
kitchoi: Should it append to the containers here anyway?  This method looks for containers with a given set of filters, it would be a surprise if the results are further filtered by the existence of ip/port.

stefanoborini: True, but the problem is that technically we can't really do much with those containers, and if that information is not present, injecting them in the rest of the code won't make them any better.

So, either we filter them at this stage, or risk to inject "validity" into a container that is already kind of broken to begin with...

I'm not sure.

remoteappmanager/docker/container_manager.py
```diff
@@ -127,8 +127,67 @@ def containers_from_mapping_id(self, user_name, mapping_id):
             'label': ['{0}={1}'.format(k, v) for k, v in labels.items()]
         }
 
+        containers = yield self.containers_from_filters(filters=filters)
+
+        return containers
+
+    @gen.coroutine
+    def container_from_url_id(self, url_id):
```
itziakos: (layout): This function does not have to be 12 lines of code. I would prefer that we use empty lines inside functions only if necessary. 

For example the following:

```
labels = {SIMPHONY_NS+"url_id": url_id,}
filters = {
       'label': ['{0}={1}'.format(k, v) for k, v in labels.items()]
}
containers = yield self.containers_from_filters(filters=filters)
return containers[0] if len(containers) else None
```

is equally understandable.

stefanoborini: ok. one sec.

PR #104, opened 2016-07-14
Fix ResourceWarning in tests

### LINKED ISSUE #60: Test report unclosed files
sys:1: ResourceWarning: unclosed file <_io.FileIO name=6 mode='wb'>
/home/travis/virtualenv/python3.4.3/lib/python3.4/importlib/_bootstrap.py:2150: ImportWarning: sys.meta_path is empty
sys:1: ResourceWarning: unclosed file <_io.FileIO name=5 mode='rb'>

Strange because my guess it's that they are used as context managers. I suspect some temporary file.

kitchoi:
TestSpawner uses ioloop and should use tornado.testing.AsyncTestCase
Fix #60

### DIFF
--- a/tests/spawner/test_spawner.py
+++ b/tests/spawner/test_spawner.py
@@ -1,40 +1,36 @@
 import contextlib
 import os
 import pwd
-import shutil
 import subprocess
 import sys
 import time
-import tempfile
-import unittest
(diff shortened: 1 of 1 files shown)


stefanoborini:
👍

Issue #54, opened 2016-07-06
Consistency with url vs urlpath


stefanoborini:
There's a general inconsistency in "url" vs "urlpath" that makes things harder to understand.

- a url is according to `scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]`, meaning that `http://host:8000/foo/bar` is a url. `/foo/bar` is not.
- a urlpath is a path that goes into a url. We want this distinction because some paths are on the physical machine instead, so just saying path does not convey the appropriate context.

This inconsistency is inherited from jupyterhub, but now that we have control of the spawner and we pass what we want at the command line, we should be clear on the intent to avoid confusion.

itziakos:
is this fixed now?

stefanoborini:
@itziakos Partially, I fear. The merged solution fixes the option, but internally I am not sure we are consistent everywhere.

kitchoi:
@sbo @itziakos shall we close this and open new issue when we spot missed loners?  I don't guarantee I cleaned up all of them, but I tried my best.

itziakos:
Yes it is fine to open new issues on this

PR #90, opened 2016-07-12
REST API [WIP]

stefanoborini:
Experimental implementation of an application resource.

PR #105, opened 2016-07-15
Rename url to urlpath where it is possible and appropriate

kitchoi:
For #54


kitchoi:
For #54

kitchoi:
Will have a small merge conflict with #90

stefanoborini:
That's a good step in the right direction, but I am also concerned about internal uses. in some places, we might use one instead of the other.

stefanoborini:
@kitchoi I merge this, so I can bring your changes into the REST.

PR #110, opened 2016-07-15
Documentation notes on HTTP connection to docker

kitchoi:
Not sure if I want to modify the binary png image just for this.


kitchoi:
For #3

itziakos:
:+1:

PR #111, opened 2016-07-15
Adds flake to make test

stefanoborini:


Makefile
```diff
@@ -56,4 +56,4 @@ images:
 test:
 	@echo "Running testsuite"
 	@echo "-----------------"
-	-. venv/bin/activate && python -m tornado.testing discover
+	-. venv/bin/activate && flake8 remoteappmanager tests && python -m tornado.testing discover
```
itziakos: why not add another target `flake8` and then `test: flake8` should do the right  thing

itziakos:
minor comment :+1:

Makefile
stefanoborini: Because I don't think typing `make flake8` is better than typing `flake8 .`. In fact, I've been excessive in that. It should be `flake8 .`, because the tox.ini will take care of excluding the irrelevant stuff.

PR #115, opened 2016-07-18
Ensure mapping id from CSVAccounting contains only hex characters

kitchoi:
Follows #80, mapping id containing '/' would cause problem with the rest api.
Added tests.


kitchoi:
@sbo Please review.

PR #120, opened 2016-07-18
Reverse proxy remove of container specific methods.

### LINKED ISSUE #114: Simplify the ReverseProxy
A reverse proxy does not really need to have remove and add container methods. All that is necessary is to map urls.

stefanoborini:
Brought the interface of the reverse proxy to its level of abstraction, removing methods concerning the container.

Fixes #114

### DIFF
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -104,7 +104,6 @@ class Application(web.Application, LoggingMixin):
         return ReverseProxy(
             endpoint_url=self.command_line_config.proxy_api_url,
             auth_token=auth_token,
-            base_urlpath=self.command_line_config.base_urlpath
         )
 
     @default("hub")
--- a/remoteappmanager/application.py
+++ b/remoteappmanager/application.py
@@ -19,6 +19,7 @@ from remoteappmanager.services.reverse_proxy import ReverseProxy
 from remoteappmanager import rest
 from remoteappmanager.rest import registry
 from remoteappmanager import restresources
+from remoteappmanager.utils import url_path_join
 
 
 class Application(web.Application, LoggingMixin):
@@ -148,6 +149,16 @@ class Application(web.Application, LoggingMixin):
         self.listen(self.command_line_config.port)
 
         tornado.ioloop.IOLoop.current().start()
+
+    def urlpath_for_object(self, object):
+        """
+        Resolves the absolute url path of a given object.
+        The object must have a urlpath property.
+        """
+
+        return url_path_join(
+            self.command_line_config.base_urlpath,
+            object.urlpath)
 
     # Private
     def _get_handlers(self):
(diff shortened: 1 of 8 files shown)


remoteappmanager/handlers/home_handler.py
```diff
@@ -112,7 +112,9 @@ def _actionhandler_start(self, options):
 
         # The server is up and running. Now contact the proxy and add
         # the container url to it.
-        urlpath = yield self.application.reverse_proxy.add_container(container)
+        urlpath = self.application.urlpath_for_object(container)
```
kitchoi: Why not `url_path_join(self.application.command_line_config.base_urlpath, container.urlpath)` here?
Note that `base_urlpath` is also retrievable in `HomeHandler.settings['base_urlpath']`

remoteappmanager/application.py
```diff
@@ -149,6 +149,16 @@ def start(self):
 
         tornado.ioloop.IOLoop.current().start()
 
+    def urlpath_for_object(self, object):
```
kitchoi: Perhaps a subjective comment, I am not sure about the merit of having one function that simply calls another function.

tests/restmodel/test_container.py
```diff
@@ -28,11 +29,10 @@ def get_app(self):
         app = web.Application(handlers=handlers)
         app.file_config = Mock()
         app.file_config.network_timeout = 5
+        app.urlpath_for_object = Mock(return_value="/urlpath_for_object/")
```
kitchoi: Does this need to be mocked?  Can we test it?

kitchoi:
Except for a few comments above (mainly regarding `Application.urlpath_for_object`), the rest looks good to me.

remoteappmanager/application.py
stefanoborini: Level of abstraction.
The point is this: we have applications, and they are mapped to urls in our server. Applications can't know what is their url, but everything that handles them must know this information.

Who knows this information (better, who is authoritative for this information)? The application. So you should ask the application. What is the url for a given application? and the application replies. Whatever it replies it's added to the reverse proxy.

remoteappmanager/handlers/home_handler.py
stefanoborini: Because there should only be one authoritative source for a given information. And with containers the container itself can't know. It can only know its own relative path, but to know the absolute one, the application is authoritative, so it should be inquired, and it should be inquired with an interface that is generic.

I am not saying this method is the perfect solution. I am saying is that we would be scattering around logic that is not the concern of the handler.

tests/restmodel/test_container.py
stefanoborini: Yes good point, we need to test it.
But this is mocked there because we are using this generic web.Application in this test.
The reason is that we can't instantiate our application. We would need to mock a lot of stuff to do so and we are not ready for that yet.

PR #121, opened 2016-07-18
Extract reference number in logging

### LINKED ISSUE #73: Formalise error handling with the logger and the refnumber for the issue
Currently when an error takes place in a number of places we generate a `refnumber` that can be displayed in the user when an error takes place. This `refnumber` is at the same time logged next to the related error. This will help to pinpoint the place in the logs where a specific error (reported by a user) took place and simplify debugging.

To achieve this I would suggest one of the following:

- Wrap the python `Logger`  logging functions so that they generate a refnumber, pass it to the logger and then return it to the caller.
- Provide a generic `log(self, level, msg, *args, **kwargs)` function in the `BaseHandler` that will generate the refnumber pass it to a python logger and return it.

I think that the second implementation is simpler and easier to test.

stefanoborini:
Isolated reference number generation in the logger. Currently used only in one place.
Additionally, removed the traits logger in favour of the tornado one.

Fixes #73.


kitchoi:
:+1:

Issue #113, opened 2016-07-15
remoteappdb creates a database when it does not exist


kitchoi:
Commands apart from `init` create a database when it does not exist.
```
$ ls
$ 
$ remoteappdb user create username
....
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: user [SQL: 'INSERT INTO user (name) VALUES (?)'] [parameters: ('username',)]
$ ls
sqlite.db
$
```

itziakos:
The right behavior here is to raise an error if the db does not exist before attempting to initialize the sqalchemy wrapper.

PR #122, opened 2016-07-18
Remoteappdb prevents run if the database is not present

stefanoborini:


kitchoi:
The `db_url` is managed by the `RemoteAppDBContext`, which already normalises the url, so `is_sqlitedb_url(db.url)` will always be True.  Alternatively one could have a `RemoteAppDBContext.db_exists` method that returns `True` if the database already exists.  This way it would be easy to extend `RemoteAppDBContext` for handling other database dialects (e.g. postgres).  Up to you, good to merge otherwise.

stefanoborini:
> which already normalises the url, so is_sqlitedb_url(db.url) will always be True

not if the url is mysql://server/

kitchoi:
> not if the url is mysql://server/
Yes I noticed that, but likewise this PR does not fix the problem for other dialects, so I understood that 'sqlite' is an existing assumption for remoteappdb, am I right?

stefanoborini:
@kitchoi Technically, if you have a mysql database, there's pretty much nothing preventing this script to work and create all information on the mysql database, with maybe the exception of login/password information.

kitchoi:
I believe this PR is fixing #113 right?  Does the issue exist for database dialects other than sqlite?

stefanoborini:
If you have a different database, it will always perform the operations regardless. What you don't want, is to use the logic in place for checking sqlite databases when you passed mysql://whatever

kitchoi:
Thanks @sbo.  After our offline discussion, I believe the issue #113 is only relevant to sqlite databases, so that is all the fix this PR needs to provide.  For other database dialects, we can rely on the admin-level control on database creation.

@sbo, it would be nice to add a comment near `if is_sqlite_url...` that mentions the file-creation behaviour on `commit`, for future reference (and for other developers).

Please merge when you're ready.  Thanks again.

Issue #98, opened 2016-07-14
Implement reasonable mock objects for testing.


stefanoborini:
At the moment, there's a lot of ad hoc patching going on to define mock for tests. We are reaching a point when it's no longer a one-shot operation, so it's probably a good idea to define mocks for some well defined, consistent objects (e.g. docker client, container manager, user, database). This will remove or reduce the need for coroutine factories that we are using right now.

itziakos:
Since we have interfaces I agree that having `virtual` or `dummy` implementations is a goo thing. Please note that I would suggest to avoid using the word `mock` for such objects to avoid confusion with the `mock` library.

itziakos:
@stefanoborini  and @kitchoi can you compile a list of the objects (please provide the class name) that we should create dummy or virtual versions for?

kitchoi:
I propose creating dummy and virtual objects for the following (I will update this message locally to avoid spams), some of which already exist:

Dummy:
- `remoteappmanager.Application`
- `remoteappmanager.file_config.FileConfig`
- `remoteappmanager.command_line_config.CommandLineConfig`
- `remoteappmanager.db.orm.AppAccounting`
- `remoteappmanager.db.csv_db.CSVAccounting`
- `remoteappmanager.services.hub.Hub`
- `remoteappmanager.docker.container_manager.ContainerManager`
- `remoteappmanager.services.reverse_proxy.ReverseProxy`

Virtual:
- `docker.Client` (from docker-py)
-  `jupyterhub.orm.Proxy`
- `request.post` and `request.get` (for testing REST api and would be needed when we use the REST api in HomeHandler)

Note on terminology (at least for me): 
- "Dummy" objects are created for the corresponding class, with mocked methods (etc), it can rely on virtual objects for its mocked methods.
- "Virtual" object itself is mocked (e.g. `Mock(spec=Class)`)

PR #125, opened 2016-07-19
Add dummy and virtual objects for testing

kitchoi:
Provides for #98 

Work in progress.
(should have named the branch differently...never mind)


kitchoi:
@sbo I have added a number of dummy/virtual objects for Application, Accounting, Hub, ContainerManager and docker.Client

I had to modify some tests to validate that these objects work.  I also noticed that `TestContainerManager` would have clash with #126, in which case I am happy to resolve the conflicts on this branch.

Some existing tests can switch to using these dummy objects, but I have not done in order to make it easier to review.

stefanoborini:
Ok, the other branch touches more files, so to reduce risk let's merge that one first and this one second. We'll do the switch progressively or in a separate PR.

In the meantime, I review this one and you can review the other one.

tests/restmodel/test_container.py
```diff
@@ -131,7 +131,7 @@ def test_retrieve(self):
             self.assertEqual(res.code, httpstatus.NOT_FOUND)
 
             self._app.container_manager.docker_client.containers = \
-                mock_coro_factory(return_value=[utils.containers_dict()])
+                mock_coro_factory(return_value=[containers_dict()])
```
stefanoborini: I wonder if we can mock the internal sync client instead...

Just a comment. I would not do it now, but useful to consider. I changed the AsyncDockerClient.client to _sync_client, so if you change it now it would need change again. Let's wait.

stefanoborini:
General comment. A lot of the current boilerplate output we have in utils is now probably ready to disappear, which is good news

tests/mocking/dummy.py
```diff
@@ -0,0 +1,138 @@
...
+        return (('12345',
+                 DummyDBApplication(image='image1'),
+                 DummyDBApplicationPolicy()),
+                ('id678',
+                 DummyDBApplication(image='image2'),
+                 DummyDBApplicationPolicy()))
+
+
+def create_reverse_proxy(params=None,
+                         register_result='',
```
stefanoborini: It would be convenient at this point to change mock_coro_factory to do exceptions as well as a return value, like mock does. It would simplify our general testing of exceptions.

stefanoborini: Again, don't work on that. It's good to keep it into account as a "small task". We can open individual issues for these.

stefanoborini:
ok. now there's a merge conflict as predicted.

kitchoi:
@sbo And it is resolved!

PR #126, opened 2016-07-19
Docker config file defaults from environment

### LINKED ISSUE #85: Docker environment variables can be used as default docker client settings
A number of environment variables are *already* available in the docker machine, including TLS settings.  Instead of requiring the remoteappmanager config file to provide *all* parameters for the docker client, we should use the environment variables when parameters are not specified in the config file.

For example, the remoteappmanager file config can provide just
```
tls_verify = ...
```
and this would override the environment variables `$DOCKER_TLS_VERIFY` while the other parameters would have default settings such as `tls_cert = $DOCKER_CERT_PATH/cert.pem` (`docker.from_env` can handle this path construction for you).

Note on implementation details: ContainerManager is configured by DockerClientConfig configurated by the config file.  If DockerClientConfig is not given, environment variables are used.  But we always provide such DockerClientConfig so the config file is always used.

Reference:
https://docs.docker.com/engine/security/https/#/secure-by-default
https://books.google.co.uk/books?id=C8XeCgAAQBAJ&pg=PA125&lpg=PA125&dq=kwargs_from_env+docker+py&source=bl&ots=pTVPUI769C&sig=rg7MkbDBZGnPvBNBFwVbszmRuf4&hl=en&sa=X&ved=0ahUKEwjj8prZv-3NAhXlIsAKHaRbAhMQ6AEIMDAD#v=onepage&q=kwargs_from_env%20docker%20py&f=false

stefanoborini:
Fixes #85


kitchoi:
I like the fact that default values for tls parameters are extracted from `docker.utils.kwargs_from_env`.  However I would prefer they be defined as traitlet's defaults so that `docker.utils.kwargs_from_env` is not triggered altogether if one provides all the required parameters in the config file.  It might be a bit repetitive but this is because `docker.utils.kwargs_from_env` automatically checks if the default cert files exist and such check might fail for exactly the same reason why one would want to set the parameters in the file config.
Ref: https://github.com/docker/docker-py/blob/1.8.1/docker/tls.py#L44
https://github.com/docker/docker-py/blob/986a14a152d20df6bb8622516ab2d18af73e5e20/docker/utils/utils.py#L523

stefanoborini:
Well, if you have the environment variables defined but the certificates are not there, you surely want an error to be raised.

kitchoi:
> Well, if you have the environment variables defined but the certificates are not there, you surely want an error to be raised.

Isn't environment variables such as `DOCKER_CERT_PATH` set by docker toolbox (and our spawner)?  So if one wants to override, e.g. the tls cert parameters, using the file config, the relevant environment variables have to be manually unset?  Is this the behaviour we want?

kitchoi:
> Isn't environment variables such as DOCKER_CERT_PATH set by docker toolbox (and our spawner)? So if one wants to override, e.g. the tls cert parameters, using the file config, the relevant environment variables have to be manually unset? Is this the behaviour we want?

Imagine a scenario where the default cert path does not have the cert files but you have cert files somewhere else.  If my mental compilation of the PR is correct, you'd have to either (1) set the environment variable `DOCKER_CERT_PATH` appropriately or (2) unset the `DOCKER_CERT_PATH` and then set the parameters in the config file or (3) put the cert files in the default location.

remoteappmanager/traitlets.py
```diff
@@ -18,10 +18,10 @@ def set_traits_from_dict(traited_instance, d):
     Note: if a set operation fails, the appropriate traitlet exception is
     raised. Traitlets that were already set won't be rolled back.
     """
-    for traitlet_name, traitlet in traited_instance.traits().items():
-        traited_instance.set_trait(
-            traitlet_name,
-            d[traitlet_name])
+    dict_keys = d.keys()
+    traitlet_names = traited_instance.traits().keys()
```
kitchoi: Same as `traited_instance.trait_names()`

kitchoi:
Regarding the comment on the non-existence of cert/key/ca.pem files in the default cert path, whatever we decide the behaviour should be, I think there should be a test for such scenario.
Apart from this, the rest looks good to merge to me.

remoteappmanager/file_config.py
```diff
@@ -57,6 +64,34 @@ class FileConfig(HasTraits):
...
+        env = os.environ
+
+        self.docker_host = env.get("DOCKER_HOST", "")
+        if self.docker_host == "":
+            self.docker_host = "unix://var/run/docker.sock"
+
+        self.tls_verify = (env.get("DOCKER_TLS_VERIFY", "") != "")
+
+        cert_path = env.get("DOCKER_CERT_PATH", "")
+        if self.tls_verify or cert_path != "":
```
kitchoi: Should we set these defaults no matter what?

remoteappmanager/file_config.py
```diff
@@ -57,6 +64,34 @@ class FileConfig(HasTraits):
...
+            if cert_path == "":
+                cert_path = os.path.join(os.path.expanduser("~"), ".docker")
+
+            self.docker_host = self.docker_host.replace('tcp://', 'https://')
```
kitchoi: Why 'https://' instead of 'tcp://'?

remoteappmanager/file_config.py
stefanoborini: No because otherwise it will enable the tls verification later on, when you generate the dict...

remoteappmanager/file_config.py
stefanoborini: That's how docker does it...

Which brings me to the point if we should replace also the config file entry if wrong, or should we just go straight as it says....

At the moment, what's in the file is used verbatim.

remoteappmanager/file_config.py
kitchoi: > when you generate the dict...

Can you point me to this dict please?  Just wanted to understand it better.

stefanoborini: docker_config, although I noticed it creates a tls section only if you have tls_verify. It should also check for certificate paths. tls_verify == False does not imply that you don't perform tls.

remoteappmanager/file_config.py
kitchoi: For the `https://` comment, I see that here now: https://github.com/docker/docker-py/blob/1.9.0-rc2/docker/utils/utils.py#L509

1. Would be good to add a comment here to remind ourselves or others in the future (because it looks like a hack)
2. ContainManager._get_ip_and_port (https://github.com/simphony/simphony-remote/blob/e3045d31926195a81833b1bb2f0f4d04a79b24e5/remoteappmanager/docker/container_manager.py#L392) checks if the hostname starts with `tcp`.  In fact we just take the host after `://` there.

remoteappmanager/docker/container_manager.py
```diff
@@ -376,8 +386,9 @@ def _get_ip_and_port(self, container_id):
         # docker url to extract the ip.
         ip = '127.0.0.1'
 
-        if self.docker_config and self.docker_config.docker_host != '':
-            url = urlparse(self.docker_config.docker_host)
+        base_url = self.docker_config.get("base_url")
+        if base_url:
+            url = urlparse(base_url)
             if url.scheme == 'tcp':
```
kitchoi: https://github.com/simphony/simphony-remote/pull/126/files#r71502734 refers to here.  Maybe we should take the hostname regardless of its scheme?

remoteappmanager/file_config.py
kitchoi: Anyhow this particular line-comment is IMO resolved.

kitchoi:
Two last minor comments, good to merge then! Thanks @sbo!

remoteappmanager/file_config.py
```diff
@@ -94,4 +132,35 @@ def parse_config(self, config_file):
...
+        params = {}
+        params["base_url"] = self.docker_host
+
+        # Note that this will throw if the certificates are not
+        # present at the specified paths.
+        # Note that the tls flag takes precedence against tls verify.
+        # This is docker behavior.
+        if self.tls:
+            params["tls"] = tls.TLSConfig(
+                client_cert=(self.tls_cert, self.tls_key),
```
kitchoi: if `tls` is True but `tls_verify` is False, `tls_cert` and `tls_key` are not set.  Why do we need cert and key when `tls` is True?

kitchoi: Note that `docker.Client(tls=True)` does completely different thing: https://github.com/docker/docker-py/blob/master/docker/client.py#L90

stefanoborini: ok, but we never have tls=True. we have it in the file config but our meaning is the same as the —tls and —tlsverify in docker as a result, we always pass a TLSConfig to the client or nothing, in that case it’s False

stefanoborini: > if tls is True but tls_verify is False, the default of tls_cert and tls_key are not set. Why do we need cert and key when tls is True?

Yes they are. They are defined on the cert_path variable. Check the init

kitchoi: I meant this:
```
In [1]: from remoteappmanager.file_config import FileConfig

In [2]: tmp = FileConfig(tls=True)

In [3]: tmp.tls_cert, tmp.tls_key
Out[3]: ('', '')

```

kitchoi:
@sbo merge?

PR #133, opened 2016-07-20
Container from dictionary agnostic of data source

### LINKED ISSUE #84: Image.from_docker_dict and Container.from_docker_dict take inconsistent inputs
`Container.from_docker_dict` assumes inputs from the result of `docker.client.containers` but not `docker.client.inspect_container`, whilst `Image.from_docker_dict` can take inputs from both.

It would be useful and neater if `Container.from_docker_dict` accepts input from the result `docker.client.inspect_container` as well.

Follows #79

stefanoborini:
Fixes #84


kitchoi:
:+1:

PR #141, opened 2016-07-21
Ensure no empty file created if target sqlite database does not exist

### LINKED ISSUE #139: Sqlite database file created if it does not exist
Similar to #113, when `remoteappmanager` is started with its default accounting setting and that the default sqlite database does not exist, an empty file `remoteappmanager.db` would be created.

kitchoi:
Fix #139

### DIFF
--- a/remoteappmanager/db/orm.py
+++ b/remoteappmanager/db/orm.py
@@ -158,6 +158,7 @@ class AppAccounting(ABCAccounting):
 
     def __init__(self, url, **kwargs):
         self.db = Database(url, **kwargs)
+        self.check_database_readable()
 
     def check_database_readable(self):
         ''' Raise IOError if the database url points to a sqlite database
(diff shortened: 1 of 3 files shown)


remoteappmanager/db/orm.py
```diff
@@ -158,10 +159,26 @@ class AppAccounting(ABCAccounting):
...
+            file_path = os.path.abspath(db_url[10:])
+            if not os.access(file_path, os.R_OK):
+                raise IOError(
+                    'Sqlite database {} is not readable'.format(file_path))
+
     def get_user_by_name(self, user_name):
         """ Return an orm.User given a user name.  Return None
         if the user name is not found in the database
         """
+        self.check_database_readable()
```
stefanoborini: The problem with this solution is that now we are performing this check for every request that is issued. I am not sure if there are other factors in play, but maybe we can just do the check at init, and then assume it's there from then onwards.

kitchoi: Yep I think that's a reasonable assumption.

PR #146, opened 2016-07-21
Activates venv at start

stefanoborini:
Prevents running without venv activated. Just a courtesy.

### DIFF
--- a/jupyterhub/start.sh
+++ b/jupyterhub/start.sh
@@ -1,1 +1,1 @@
-jupyterhub --ssl-key test.key --ssl-cert test.crt
+. ../venv/bin/activate && jupyterhub --ssl-key test.key --ssl-cert test.crt


kitchoi:
I don't think this is necessary... for example, the user's virtual environment may not be called `venv`

stefanoborini:
we create it as such in the makefile. @kitchoi

kitchoi:
IMHO the start script is in fact not needed, to keep the directory slim.  If we need to provide such shortcut and that shortcut has to be consistent with the Makefile, a 'start' target can be provided inside the Makefile.  Outside of it, the script should be general.  Alternatively we can update the deployment documentation.

kitchoi:
I don't have a strong opinion afterall.  I personally don't use this script.

kitchoi:
@sbo will this break the travis's yaml in #150?

PR #147, opened 2016-07-21
Update documentation on setting up database, accounting and spawner

kitchoi:


remoteappmanager/file_config.py
```diff
@@ -16,13 +16,13 @@ class FileConfig(HasTraits):
     ##########
     # Configuration file options. All of these come from the config file.
 
-    #: Enable tls, with a twist. if we use self-signed certificates,
-    #: using tls as True will produce an error of incorrect CA validation.
-    #: As a consequence, defaults to False. TLS secure connection will still
-    #: happen thanks to tls_verify and tls[_cert|_key|_ca] being defined.
-    #: See https://docs.docker.com/engine/security/https/
-    #: Verification can be enabled simply by issuing tls=True in the
-    #: config file
```
kitchoi: @sbo I am not sure, is this meant to be on the public doc?

stefanoborini: Yes, because the meaning of those keys is tricky. We want to refer to the appropriate info.

kitchoi:
@sbo can we merge this for the release?

Issue #136, opened 2016-07-20
Clean up tests.utils now replaceable by dummy and virtual objects


kitchoi:
Following #125, a number of functions in `tests.utils` can be cleaned up.

They are, not exhaustive:
`tests.utils.mock_docker_client*`
`tests.utils.container_dicts`

kitchoi:
close via #172

PR #153, opened 2016-07-21
Switch to use virtual docker client in some tests

kitchoi:
Follows #125

Fixed bugs in the virtual docker client's `containers` method
(see tests/mocking/virtual/docker_client.py)


kitchoi:
Relevant for #136

Issue #143, opened 2016-07-21
Consider setting a default Accounting to something that always works


kitchoi:
Currently the default Accounting for `remoteappmanager` uses a sqlite3 database.  The Accounting would fail when the database is not setup.

This makes the application "not-work-out-of-the-box".

Potential default Accounting class may be one that returns all docker images in the host machine?  Or just an empty list of application?

stefanoborini:
What we can do is to indeed have a "world accounting" that exports all images to all users, but it opens up one problem: distinguish images that are "valid" for us from images that aren't.

So we need to baptise a label to recognise an image as "one of us". Right now we don't really have such label, and taking advantage of any one of the currently present labels would be risky.

Any other strategy could open up to false positives (e.g. checking if only one port is exported, or if it derives from a specific base image)...

kitchoi:
I personally prefer an empty list of applications then touching docker, higher chance of it to 'always work'

itziakos:
I prefer that the spawner caprture's the error and shows the user an empty page with suggestions on how to fix the issue. ( e.g. link to the troubleshoting section in the docs)

kitchoi:
Close via #159 and #160

PR #159, opened 2016-07-25
Add custom error page for uncaught exceptions

kitchoi:
Uncaught error such as #143 causes a plain "HTTP 500: Internal Server Error" page.  This PR overloads tornado's [`write_error`](http://www.tornadoweb.org/en/stable/web.html?highlight=write_error#tornado.web.RequestHandler.write_error) so that error message is displayed.

For #143, it displays something like this:
screenshot from 2016-07-26 11 29 31


stefanoborini:
I'm not sure it's a good idea to leak internal information that way. I would leave the details for the internal log, but stay generic for the presented page.

kitchoi:
This error page captures all uncaught exception, any generic exception.  If the content of the exception is not displayed, then it is no different from the original behaviour of a plain 500 server error page.

stefanoborini:
and that's the point. You don't want to leak internal information to externals. That way anyone can see exactly what went wrong and the path information. It is information leakage http://projects.webappsec.org/w/page/13246936/Information%20Leakage

kitchoi:
@stefanoborini better?
screenshot from 2016-07-26 11 29 31

kitchoi:
@stefanoborini I know this is not the best place for an HTTPError, but I would prefer improving on this after the PR list is shorter.

kitchoi:
And this is how it looks to the admin
```
[I 160726 12:06:00 orm:131] Creating session to db: sqlite:///remoteappmanager.db
[E 160726 12:06:00 application:118] The database is not initialised properly.
      File "/home/kit/.virtualenv/simremote/lib/python3.4/site-packages/traitlets/traitlets.py", line 501, in get
        value = obj._trait_values[self.name]
    KeyError: 'user'
    
    During handling of the above exception, another exception occurred:
    
      File "/home/kit/.virtualenv/simremote/lib/python3.4/site-packages/traitlets/traitlets.py", line 501, in get
        value = obj._trait_values[self.name]
    KeyError: 'db'
    
    During handling of the above exception, another exception occurred:
    
      File "/mnt/hgfs/VM_shared/SimPhoNy/s2imphony-remote/remoteappmanager/application.py", line 115, in _db_default
        return cls(**self.file_config.accounting_kwargs)
      File "/mnt/hgfs/VM_shared/SimPhoNy/s2imphony-remote/remoteappmanager/db/orm.py", line 176, in __init__
        self.check_database_readable()
      File "/mnt/hgfs/VM_shared/SimPhoNy/s2imphony-remote/remoteappmanager/db/orm.py", line 190, in check_database_readable
        'Sqlite database {} is not readable'.format(file_path))
    OSError: Sqlite database /mnt/hgfs/VM_shared/SimPhoNy/s2imphony-remote/jupyterhub/remoteappmanager.db is not readable
[E 160726 12:06:00 logging_mixin:26] Internal Server Error : HTTP 500: The database is not initialised properly.. Ref: efc99ca8-5320-11e6-b741-000c29ae0276

```

tests/handlers/test_base_handler.py
```diff
@@ -0,0 +1,85 @@
...
+                         )
+
+        self.assertEqual(res.code, 500)
+        self.assertIn('Internal Server Error', str(res.body))
+        self.assertIn(" Ref.:", str(res.body))
+
+
+class TestBaseHandlerDatabaseError(TempMixin, utils.AsyncHTTPTestCase):
+    def setUp(self):
+        self._old_proxy_api_token = os.environ.get("PROXY_API_TOKEN", None)
```
stefanoborini: This code shows up again and again in all our tests. It's becoming a liability.

remoteappmanager/handlers/base_handler.py
```diff
@@ -40,3 +41,20 @@ def render(self, template_name, **kwargs):
...
+        message = ""
+
+        # If this error was caused by an uncaught exception
+        # log exception message and reference number as well
+        exc_info = kwargs.get('exc_info')
+        if exc_info:
+            exception = exc_info[1]
+            ref = self.log.issue(status_message, exception)
+            reason = getattr(exception, 'reason', '')
+            message = '{} Ref.: {}'.format(reason, ref)
```
stefanoborini: You still open up to the chance that this code is actually presenting the user with internal information. It depends on the exception reason text, and if the exception text contains sensitive information for whatever reason, it will be shown to the user.

kitchoi: By default, `web.HTTPError.reason` is empty.  If it is empty, the `str(HTTPError)` representation is deduced in by the status code (e.g. 404 -> Not Found) .  If someone raises an HTTPError with a `reason` attribute, that content would be displayed to the user *even if the `write_error`* was not overload.  So it is entirely the web developer's choice what to put in `reason`.
screenshot from 2016-07-27 17 15 19

stefanoborini: which brings us back to the `reason` discussion. Is it "proper" to change the HTTP reason, when it's supposed to be a standard string? What are others doing? Tornado documentation says that reason is specified only when using non standard codes, not that we can put anything in it.

I am not saying it's not possible, just if it's good practice, and from what I know of HTTP, it isn't

kitchoi: w3.org says the reason phrases are only recommended, they may be replaced.

kitchoi: If renaming`reason` to something e.g.`user_message` makes the usage here more reasonable, then it is not the problem of the usage, it is a problem with naming.

stefanoborini: Tornado, on the other hand, seems to favour the `reason` usage only if you use non-standard codes.

reason (string) – Keyword-only argument. The HTTP “reason” phrase to pass in the status line along with status_code. Normally determined automatically from status_code, but can be used to use a non-standard numeric code.

the RFC says that it's for human consumption, but it doesn't say the scope of this information, or if it's good practice to change it to whatever 

The individual values of the numeric status codes defined for HTTP/1.1, and an example set of corresponding Reason-Phrase's, are presented below. The reason phrases listed here are only recommendations -- they MAY be replaced by local equivalents without affecting the protocol.

It's a MAY, which means it's everybody's game. So for the technical point of view it's fine. The question is if it's an acceptable practice _for_ _us_. @itziakos we need your two cents?

kitchoi: > Tornado, on the other hand, seems to favour the reason usage only if you use non-standard codes.
In their documentation, by 'non-standard codes', they mean 'non-standard status code' and therefore no standard reason phrase can be found (say a status code 420).  In many places in their own code, they custom the 'reason' for common status codes such as 403. (See `StaticFilePathHandler. validate_absolute_path`)

kitchoi: > In their documentation, by 'non-standard codes', they mean 'non-standard status code' and therefore no standard reason phrase can be found (say a status code 420). 

That said, they mean to suggest using `reason` to support non-standard status code.  I do not yet see the implication of 'don't change reason if your status code is a standard one'

kitchoi: For reference, the following is the tornado's documentation we both refer to:
> reason: Keyword-only argument.  The HTTP "reason" phrase
        to pass in the status line along with ``status_code``.  Normally
        determined automatically from ``status_code``, but can be used
        to use a non-standard numeric code.

tests/handlers/test_base_handler.py
itziakos: It is always tricky to have to cleanup/recover the enviroment variables. Can you open an issue, this will need to be investigated.

stefanoborini: It's already there.

remoteappmanager/handlers/base_handler.py
itziakos: After an offline discussion it looks that there is no obvious way of doing this. So this is a good solution for us and we will keep an eye for a better solution.

Issue #130, opened 2016-07-20
Tests on remoteappdb using number of lines in output is not reliable


kitchoi:
For example `tests.cli.test_remoteappdb.TestRemoteAppDbCLI.test_app_revoke` checks if stdout only has got 3 lines fail on occasion like this

(I did a `print(out)` in the test)
```
test_app_revoke (tests.cli.test_remoteappdb.TestRemoteAppDbCLI) ... [I 160720 14:22:56 orm:130] Creating session to db: sqlite:////tmp/tmp1aec8owd/test.db
[I 160720 14:22:56 orm:130] Creating session to db: sqlite:////tmp/tmp1aec8owd/test.db
[I 160720 14:22:56 orm:130] Creating session to db: sqlite:////tmp/tmp1aec8owd/test.db
[I 160720 14:22:56 orm:130] Creating session to db: sqlite:////tmp/tmp1aec8owd/test.db
[I 160720 14:22:56 orm:130] Creating session to db: sqlite:////tmp/tmp1aec8owd/test.db
[I 160720 14:22:57 orm:130] Creating session to db: sqlite:////tmp/tmp1aec8owd/test.db
/home/kit/.virtualenv/simremote/lib/python3.4/site-packages/SQLAlchemy-1.1.0b1-py3.4-linux-x86_64.egg/sqlalchemy/sql/visitors.py:309: ResourceWarning: unclosed <socket.socket fd=4, family=AddressFamily.AF_UNIX, type=2049, proto=0, raddr=/var/run/docker.sock>
  stop_on = set([id(x) for x in opts.get('stop_on', [])])
/home/kit/.virtualenv/simremote/lib/python3.4/site-packages/SQLAlchemy-1.1.0b1-py3.4-linux-x86_64.egg/sqlalchemy/sql/visitors.py:309: ResourceWarning: unclosed <socket.socket fd=5, family=AddressFamily.AF_UNIX, type=2049, proto=0, raddr=/var/run/docker.sock>
  stop_on = set([id(x) for x in opts.get('stop_on', [])])
1  user  myapp  0  0  0
         myapp  0  1  0

FAIL
```

stefanoborini:
What's the actual difference, and the underlying reason for it (if visible)?

kitchoi:
The expected output is 
```
1  user  myapp  0  0  0
         myapp  0  1  0

```
where the actual output is
```
[...]

PR #161, opened 2016-07-25
Make sure remoteappdb closes the session on exit

kitchoi:
Fix the root cause of #130


kitchoi:
Depends on #159 for an unrelated flake8 fix

Issue #140, opened 2016-07-21
Remove duplication between REST and home handler


stefanoborini:
Now that container creation is performed by REST, there's a lot of duplication between the home handler and the REST api, notably in the POST strategy, the existence of support methods and so on.

Ideally, this duplication needs to be removed by issuing REST requests from a javascript operation directly to the REST api.

stefanoborini:
On this regard, I am tempted to say that we need a supervisor object that manages the whole operation of starting and stopping a container. The container manager is, by design, focused _exclusively_ on docker, but starting a container for user consumption requires a few additional steps: setting up the reverse proxy, waiting for the service to be ready etc. Right now this logic resides in the controller class we have: the handler, but the handler is very web oriented.

I propose to:
- [ ] put most of the non-web logic in this separate "supervisor" object
- [ ] use the rest api to invoke its operations
- [ ] remove all logic from the home handler, especially the POST handler.
- [ ] leave home handler with a GET method returning a rendered template containing, in addition to the web page, a javascript rest client that can perform an authenticated rest request to our rest api, and change the buttons dynamically.

stefanoborini:
I think we should use this one https://github.com/marmelab/restful.js

stefanoborini:
Also this one https://www.npmjs.com/package/rest-js

stefanoborini:
After much pain, I decided to give up on npm/browserify/rest-js. Too many issues. I'll try with bower and restful.js, and if that doesn't work either I'll just go jquery and suffer in silence.

stefanoborini:
Giving up on bower and restful.js as well. It is unclear how bower creates the packages, and how we can integrate them in our js code. Requires webpack but generates a lot of errors due to missing packages (babel-core etc). Despite installing, could not reach a stable point.

Giving up. Going for the jquery approach.

kitchoi:
Close via #165

PR #165, opened 2016-07-26
REST JavaScript client

stefanoborini:
Uses the REST interface to start and stop containers, removing the need for the POST method in HomeHandler.


stefanoborini:
Fixes #140

remoteappmanager/restresources/container.py
```diff
@@ -4,11 +4,11 @@
 from tornado import gen
 
 from remoteappmanager.docker.docker_labels import SIMPHONY_NS
-from remoteappmanager.handlers.home_handler import _wait_for_http_server_2xx
 from remoteappmanager.rest import exceptions
 from remoteappmanager.rest.resource import Resource
 from remoteappmanager.docker.container import Container as DockerContainer
 from remoteappmanager.utils import url_path_join
+from remoteappmanager.netutils import wait_for_http_server_2xx
```
kitchoi: Could `wait_for_http_server_2xx` be placed in `utils`?

stefanoborini: I prefer it in a separate module that is aimed at network.

kitchoi:
Note that in the HomeHandler on master, if 'wait_for_http_server_2xx' throws a Timeout error, an error message is displayed and the container is removed.  It is no longer the case here, instead the start button "loads" forever.

Original behaviour (try starting 'novnc-never-starts':
screenshot from 2016-07-27 11 05 20

Behaviour for this PR:
screenshot from 2016-07-27 10 54 59

tests/test_netutils.py
```diff
@@ -0,0 +1,54 @@
...
+class TestUtils(AsyncHTTPTestCase):
+    def get_app(self):
+
+        app = web.Application(handlers=[('/short', ShortHandler),
+                                        ('/long', LongHandler)])
+        return app
+
+    @gen_test
+    def test_basic(self):
+        yield wait_for_http_server_2xx(self.get_url("/short"), timeout=2)
```
kitchoi: how to make sure the request would finish within 2 seconds?

kitchoi: Never mind that, the computer has to be reaaaaally slow to not have completed the request.

kitchoi:
There is a retrogression on the error handling (see above), but I think we can fix that on a separate PR, all within release 0.6.0?  The rest looks good to merge to me

stefanoborini:
@kitchoi  No I fix it here.

stefanoborini:
Ok, the reason why you are not seeing the error is because we are missing bootstrap. That produces an error in the javascript execution.

In any case, I ported the server side error handling (so it behaves like the old code) but I also need to introduce the javascript error handling to remove the spinner.

kitchoi:
:+1:

PR #170, opened 2016-07-26
File Config is no longer a requirement for remoteappmanager

### LINKED ISSUE #138: Config file path for remoteappmanager can now be optional
Since the default remoteappmanager config file is now basically empty, we might as well default it to None and not parse it altogether.  This would eliminate the need of creating an empty config file.

Enhancement for user experience

kitchoi:
Fix #138


kitchoi:
Need to update the doc for this change.

tests/spawner/test_spawner.py
```diff
@@ -85,13 +85,17 @@ def test_args(self):
         self.assertIn("--config-file={}".format(path), args)
         self.assertIn("--base-urlpath=/", args)
 
+    def test_args_without_config_file_path(self):
+        args = self.spawner.get_args()
+        self.assertIn("--proxy-api-url=http://127.0.0.1:12345/foo/bar/", args)
+        self.assertFalse(any("--config-file=" in arg for arg in args))
```
stefanoborini: or you could "config-file" in " ".join(args)

kitchoi: Yes, but I think that loops over the content of`args` twice?  `any` returns `True` for the first `True` (although if the test should pass, it loops over the whole thing, once).

stefanoborini: Are you seriously talking performance on a test that checks if a string is present in a 100 bytes string?

kitchoi: 😈

PR #171, opened 2016-07-26
FileConfig consistency with dockerpy

### LINKED ISSUE #149: Our FileConfig may not be consistent with the docker config
Referring to http://docker-py.readthedocs.io/en/latest/tls/

If one wants to use TLS but with self-signed certificate, docker `tls` should be set to `True` but `tls_verify` set to `False`.  Whilst in our FileConfig, we would set`tls`to `False` and `tls_verify` to `True` (and then we assumed a CA file provide in the cert path so we verify against ourselves.)

stefanoborini:
Fixes #149

### DIFF
--- a/remoteappmanager/file_config.py
+++ b/remoteappmanager/file_config.py
@@ -15,19 +15,12 @@ class FileConfig(HasTraits):
 
     ##########
     # Configuration file options. All of these come from the config file.
-
-    #: Enable tls, with a twist. If self-signed certificates are used,
-    #: setting tls as True will produce an error of incorrect CA validation.
-    #: As a consequence, defaults to False. TLS secure connection will still
-    #: happen thanks to tls_verify and tls[_cert|_key|_ca] being defined.
-    #: See https://docs.docker.com/engine/security/https/
-    #: Verification can be enabled simply by issuing tls=True in the
-    #: config file
     tls = Bool(False,
-               help="If True, connect to docker with --tls")
+               help="If True, connect to docker with tls")
 
-    tls_verify = Bool(False,
-                      help="If True, connect to docker with --tlsverify")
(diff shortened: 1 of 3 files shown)


remoteappmanager/file_config.py
```diff
@@ -77,18 +70,18 @@ def __init__(self, *args, **kwargs):
             self.docker_host = "unix://var/run/docker.sock"
 
         self.tls_verify = (env.get("DOCKER_TLS_VERIFY", "") != "")
```
kitchoi: `'DOCKER_TLS_VERIFY=0` would evaluate to `True` here...

kitchoi:
Looks good to merge.  The single comment is orthogonal to this PR, so could be fixed separately.

remoteappmanager/file_config.py
stefanoborini: DOCKER_TLS_VERIFY is not a 1=true 0=false . It's a "defined" vs "not defined" flag.

https://github.com/docker/docker/issues/22411

kitchoi: I see.  How strange.

kitchoi: It looks to me something that future developers might unknowingly change it.
Maybe adding docker/docker#22411 to the code as comment.  We don't need to do it here, we can log this as one of the auditing issues.

stefanoborini: Not really strange, it's quite common to test for definedness, and docker is not the only one doing that. Putting 1 is just a convention. It could be anything.

kitchoi: Nevertheless, the referenced issue on docker and its referenced issues show that I am not the only one surprised by this ;)

PR #186, opened 2016-07-28
Error payload

### LINKED ISSUE #175: Introduce REST error payload delivery
The current design of the REST handler does not deliver error payload.
This is needed now because we want to report REST exceptions with more detail.

stefanoborini:
Fixes #175 

Missing tests. Will be added.


stefanoborini:
Note that it's open to discussion if we should deliver a payloaded HTTPError for any exception that may occur. Right now, only REST exceptions produce an error that is potentially payloaded. Any other exception will produce a regular internal server error with no payload.

doc/source/api/remoteappmanager.db.rst
```diff
@@ -37,6 +34,5 @@ Module contents
 
 .. automodule:: remoteappmanager.db
     :members:
-    :special-members: __init__
```
kitchoi: These changes on the doc do not seem relevant to this PR, can you create an issue separately please?

kitchoi:
Alternative to catching RESTException and convert the exception into a custom `PayloadedHTTPError`, RESTBaseHandler.write_error can recognise that the Exception is a RESTException and compose the payload accordingly.

remoteappmanager/rest/rest_handler.py
```diff
@@ -25,6 +26,44 @@ def get_resource_handler_or_404(self, collection_name):
...
+            self.set_header('Content-Type', exc.content_type)
+            self.finish(exc.payload)
+        else:
+            # For non-payloaded http errors or any other exception
+            # we don't want to return anything as payload.
+            # The error code is enough.
+            self.clear_header('Content-Type')
+            self.finish()
+
+    def rest_to_http_exception(self, rest_exc):
```
kitchoi: Class method for PayloadedHTTPError?

remoteappmanager/rest/rest_handler.py
```diff
@@ -25,6 +26,44 @@ def get_resource_handler_or_404(self, collection_name):
...
+        """
+        exc_info = kwargs.get("exc_info")
+
+        if exc_info is None:
+            self.clear_header('Content-Type')
+            self.finish()
+
+        exc = exc_info[1]
+
+        if isinstance(exc, PayloadedHTTPError) and exc.payload is not None:
```
kitchoi: Alternatively, convert `RestException` to `PayloadedHTTPError` here, this way the Resource only had to raise a RestException and the payload is handled here.

remoteappmanager/rest/rest_handler.py
```diff
@@ -25,6 +26,44 @@ def get_resource_handler_or_404(self, collection_name):
...
+            self.set_header('Content-Type', exc.content_type)
+            self.finish(exc.payload)
+        else:
+            # For non-payloaded http errors or any other exception
+            # we don't want to return anything as payload.
+            # The error code is enough.
+            self.clear_header('Content-Type')
+            self.finish()
```
kitchoi: If I am correct, we need to call `finish` with the status code here, otherwise all the `raise web.HTTPError` would not have the status code included in the response. 
See http://www.tornadoweb.org/en/stable/_modules/tornado/web.html#RequestHandler.write_error

remoteappmanager/rest/rest_handler.py
```diff
@@ -25,6 +26,44 @@ def get_resource_handler_or_404(self, collection_name):
...
             raise web.HTTPError(httpstatus.NOT_FOUND)
 
+    def write_error(self, status_code, **kwargs):
+        """Provides appropriate payload to the response in case of error.
+        """
+        exc_info = kwargs.get("exc_info")
+
+        if exc_info is None:
+            self.clear_header('Content-Type')
+            self.finish()
```
kitchoi: finish with the `status_code` (the argument) as content.  `exc_info` may not be available if `web.RequestHandler.send_error` is manually called.

remoteappmanager/rest/rest_handler.py
stefanoborini: No. In the future, the conversion will be done by a serializer that may produce xml instead of json. This serializer will be a instance var.

remoteappmanager/rest/rest_handler.py
stefanoborini: This is part of the design decision. My vision is that this method only handles http-level exception types, with the conversion of exception type done at the individual http method handlers. This allows us to take appropriate action and, for example log contextually to the method handler. Converting here we lose this specificity. 

I don't know which one is best. For sure, once we were to go for a write_error method that handles rest exceptions, then it should handle all exceptions. Plus, write_error is just responsible for the payload. We would have to reimplement send_error too, so that proper status can be set.

remoteappmanager/rest/rest_handler.py
stefanoborini: finish does not take the status code.

remoteappmanager/rest/rest_handler.py
stefanoborini: the status code is set somewhere else. write error is just in charge of rendering the payload.

remoteappmanager/rest/rest_handler.py
kitchoi: > My vision is that this method only handles http-level exception types, with the conversion of exception type done at the individual http method handlers.

We can't prevent non-http-level exception from arriving at write_error.  write_error is intended to be able to handle BaseException in general.

>  This allows us to take appropriate action and, for example log contextually to the method handler. Converting here we lose this specificity.

I don't think converting a RestException to PayloadedHTTPError here stop you from logging contextually.  If you must log, you can do so in Rest*Handler
```
try:
   yield method()
except Exception:
   log.exception
   raise
```

stefanoborini: write an issue

kitchoi: Do you mean opening an issue for logging?  I only meant to say that 'converting a RestException to PayloadedHTTPError does not stop you from logging contextually', I don't mean that there is an issue with logging.

stefanoborini: If we want to handle exceptions at the write_error level I need to do some adjustments to the current code. My main concern is if it's a good idea to handle a broad range of exceptions types there.

remoteappmanager/rest/rest_handler.py
kitchoi: Sorry I meant to have the status code in the response body as text, not the response.status_code attribute.

kitchoi: It is not necessary but convenient to see the status code displayed on the page.

remoteappmanager/rest/http/payloaded_http_error.py
```diff
@@ -0,0 +1,36 @@
...
+
+        if payload is not None:
+            if not isinstance(payload, str):
+                raise ValueError("payload must be a string.")
+
+            if content_type is None:
+                content_type = "text/plain"
+        else:
+            if content_type is not None:
+                raise ValueError("Content type specified, but no payload")
```
kitchoi: Why is this a problem?

remoteappmanager/rest/rest_handler.py
```diff
@@ -25,6 +26,44 @@ def get_resource_handler_or_404(self, collection_name):
         except KeyError:
             raise web.HTTPError(httpstatus.NOT_FOUND)
 
+    def write_error(self, status_code, **kwargs):
+        """Provides appropriate payload to the response in case of error.
+        """
+        exc_info = kwargs.get("exc_info")
+
+        if exc_info is None:
+            self.clear_header('Content-Type')
```
kitchoi: 'Content-Type' is by default not set unless you overload `set_default_header`.  Before calling `write_error`, `send_error` reset the header using `set_default_header`.  Why do we need to clear 'content-type' here? (no effect if `set_default_header` is passed, override `set_default_header` if it was defined for 'content-type')

remoteappmanager/rest/http/payloaded_http_error.py
stefanoborini: Why would anyone say "application/json" and then provide no payload?

remoteappmanager/rest/rest_handler.py
stefanoborini: write error is responsible for producing the right output. The content type depends on the payload. We know the payload here, so we zero the content-type here.

remoteappmanager/rest/http/payloaded_http_error.py
kitchoi: It is not what you would expect, but does it actually cause problem? If not, seem to be an unnecessary potential failure here.

stefanoborini: It is not unnecessary. The class constraints are clear. if you specify the payload, you can or cannot specify the content type. If you don't, it's by default text/plain. If you don't specify a payload, there's no reason for specifying the content type, and if you do, you are doing something very wrong, because it means that your payload is None by accident.

kitchoi: Is it wrong to set the header content-type to something but provide an empty content?
(by the way you could default `payload` to an empty string, just an idea)

stefanoborini: There's a difference between empty payload an no payload. some http responses don't want payload, and it's invalid if they do provide one (e.g. 204 No content, or 404 Not Found)

stefanoborini: At least in the context of REST, I mean. 404 can have a payload in non-rest context

kitchoi: No payload but has content-type header - is it just because it is **wrong** in terms of definition or does it cause error?

stefanoborini: it causes an error.

kitchoi: Would you explain how please?  I just wanted to understand.

kitchoi:
Just a comment, instead of raising an Error, one could also call `send_error` and pass keyword arguments to it, the keyword arguments can then be passed to `write_error`.  These two methods are designed to be extendable.  So for example, one could instead of raising `PayloadedHTTPError`, call `send_error(status_code, ..., payload='...')`

itziakos:
:+1:

PR #251, opened 2016-09-20
Removing orm.apps_for_user

### LINKED ISSUE #211: Remove ORM apps_for_user
This function is no longer relevant. Its contents should go into the get_apps_for_user method in the AppAccounting. The various places where it's used should be converted.

stefanoborini:
Fixes #211


stefanoborini:
See #211 for rejection reasons

Issue #213, opened 2016-09-05
start_container incorrectly returns if two containers with same image are requested


stefanoborini:
in `ContainerManager.start_container`, the start_pending set that forces a return if the image is still in progress does not consider the parameters. It is technically possible to start two containers from the same image but with different parameters (e.g. one with volume mounted the other without. they will appear as different entries in the application selection page).

With the current code, the result will be to terminate the second container start for no reason at all, in some particular (and rare) conditions (e.g. user too quick starting containers)

PR #279, opened 2016-09-29
Introduced error in case of double request.

stefanoborini:
Fixes return strategy in case of double request. Now raises an exception (which is correctly handled at the webapi level) instead of returning None


stefanoborini:
Fixes #213

PR #297, opened 2016-10-06
Revert "Support for admin user"

stefanoborini:
Reverts simphony/simphony-remote#296


stefanoborini:
Reverting. Using different strategy.

Issue #284, opened 2016-09-30
Administrative web interface


stefanoborini:
It would be nice to have a web interface for administrators, so that we can do what we can do with remoteappdb via this interface.

PR #306, opened 2016-10-10
Admin application - Read only

stefanoborini:
Basic read only administrative application that is spawned for admin accounts.


stefanoborini:
WIP as I still want to do a few cleanups

stefanoborini:
Partially addresses #284

PR #352, opened 2017-03-06
Fix deps build

stefanoborini:
Fixes changes in dependencies from ubuntu trusty since the last build.


stefanoborini:
Merging. codecov complaining is irrelevant as we are not changing any python code.

PR #379, opened 2017-03-28
Migrating and isolating google analytics to tolerate failure to load

### LINKED ISSUE #369: Application list will not load if analytics.js is blocked by adblocker/tracking detection extension
If analytics.js is blocked by any extension, the whole javascript mechanism will not work. What we have to do is to load analytics.js not as a async module via requirejs, but synchronously via a normal script tag. In the code that uses its API, if the API is not there we will have to create a mock object and let it perform requests to google analytics into the void.

stefanoborini:
Fixes #369

### DIFF
--- a/jstests/tests/test_analytics.js
+++ b/jstests/tests/test_analytics.js
@@ -1,6 +1,6 @@
 define(function(require) {
     "use strict";
-    var analytics = require("analytics");
+    var gamodule = require("gamodule");
     
     QUnit.module("Google Analytics");
     QUnit.test("test without analytics", function (assert) {
@@ -9,7 +9,8 @@ define(function(require) {
             result[0] = [cmd, id, auto];
         };
         window.apidata.analytics = undefined;
-        var ga = analytics.init();
+        var ga = gamodule.init();
+        assert.notEqual(ga, undefined);
         assert.equal(result.length, 0);
     });
     
@@ -19,7 +20,8 @@ define(function(require) {
             result[0] = [cmd, id, auto];
         };
         window.apidata.analytics = {"tracking_id": "X"};
-        var ga = analytics.init();
+        var ga = gamodule.init();
+        assert.notEqual(ga, undefined);
         assert.equal(result[0][0], "create");
         assert.equal(result[0][1], "X");
         assert.equal(result[0][2], "auto");
--- a/jstests/tests/test_analytics.js
+++ b/jstests/tests/test_analytics.js
@@ -4,14 +4,24 @@ define(function(require) {
     
     QUnit.module("Google Analytics");
     QUnit.test("test without analytics", function (assert) {
+        var result=[];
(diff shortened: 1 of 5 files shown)


remoteappmanager/templates/page.html
```diff
@@ -41,6 +41,18 @@
...
 
   {% block script %}
+    <script type="text/javascript">
+      // This will always be executed and create the basic ga function that
+      // does nothing
+      window.ga = function () {
+        (window.ga.q = window.ga.q || []).push(arguments);
+      };
+      window.ga.l = 1 * new Date();
+      window.GoogleAnalyticsObject = "ga";
```
stefanoborini: There's a problem here, see if you can spot it.

stefanoborini:
@martinRenou There's a small problem in this PR, see if you can spot it

jstests/tests/test_analytics.js
```diff
@@ -4,14 +4,24 @@ define(function(require) {
     
     QUnit.module("Google Analytics");
     QUnit.test("test without analytics", function (assert) {
+        var result=[];
+        window.ga = function(cmd, id, auto) {
+            result[0] = [cmd, id, auto];
+        };
         window.apidata.analytics = undefined;
         var ga = analytics.init();
```
martinRenou: You're not using `ga` variable after that, just `analytics.init();` is enough I think.

jstests/tests/test_analytics.js
```diff
@@ -4,14 +4,24 @@ define(function(require) {
...
+        assert.equal(result.length, 0);
     });
     
     QUnit.test("test with analytics", function (assert) {
+        var result=[];
+        window.ga = function(cmd, id, auto) {
+            result[0] = [cmd, id, auto];
+        };
         window.apidata.analytics = {"tracking_id": "X"};
         var ga = analytics.init();
```
martinRenou: Same here

remoteappmanager/static/js/gamodule.js
```diff
@@ -0,0 +1,26 @@
...
+], function () {
+    "use strict";
+
+    function init() {
+        var module;
+        
+        if (window.apidata.analytics !== undefined) {
+            window.ga('create', window.apidata.analytics.tracking_id, 'auto');
+        } else {
+            window.ga = function () {
```
martinRenou: I think `window.ga = undefined;` and `window.ga = function () {};` will have the same effect but for the second option jslint would complain about empty brackets

stefanoborini: pycharm complains for the exact opposite reason. It claims return is unnecessary.
by the way, I use jshint, not jslint (jslint is the one by douglas crockford, and while I respect the guy, I think its linter is a bit excessive)

martinRenou: Ahah ok :)

PR #383, opened 2017-03-29
Make stop icon more visible [WIP]

### LINKED ISSUE #376: Make stop icon more visible
Currently we stop a running application by clicking on the icon after hovering it. This is kind of deceiving (users may instinctively click on the icon to focus) and not discoverable (you have to hover the icon to see the X). We need to find a better way to make it more evident what to click, without being too bold or in the way.

martinRenou:
Closes #376


stefanoborini:
It's not only about changing the color. I think this is still work in progress.

stefanoborini:
No longer relevant, as we decided for a different approach: put the name and the close button on top of the bar.

PR #384, opened 2017-03-30
Fix for loss of registration of reverse proxying [WIP]

stefanoborini:


stefanoborini:
Moving to something else. This is complicated by the fact that our tests actually try to spawn the remoteappmanager, and just because no request is actually performed, it just turns out fine. If we, however, start performing requests immediately after startup, things go wrong.

To address this in the proper way, we need to solve the test issues, e.g. we need a mock of the reverse proxy (not mock class, mock service) and we need to solve the problem of the database, that must be always properly setup, which may force us to have the configuration file, always.

stefanoborini:
Reworked with old strategy in  370-stopping-loses-access-2

PR #396, opened 2017-04-10
Pinning against specific version of Jupyterhub to fix red.

stefanoborini:
Apparently something has changed upstream in jupyterhub so that it makes our master red.
I pinned against 2d1a45f0190059ef436c2f97dc8d6e391eb2d139 for now, but the problem needs to be investigated further.


stefanoborini:
Error due to missing coverage. Good to go.

stefanoborini:
Further kept track on pr #396

PR #398, opened 2017-04-11
Create ember-cli webclient subproject

martinRenou:


.travis.yml
```diff
@@ -8,15 +8,25 @@ dist: trusty
...
+node_js:
+    - "6"
+
 before_install:
     - firefox --version
     - make deps
     - make pythondeps
     - python -c "import setuptools; print(setuptools.__version__)"
     - pip --version
+    - npm install -g npm
```
stefanoborini: don't install with -g. You want it local

.travis.yml
```diff
@@ -8,15 +8,25 @@ dist: trusty
...
+    - npm config set spin false
+    - npm install -g bower phantomjs-prebuilt
 
 install:
     - make install
+    - pushd webclient
```
stefanoborini: this is all makefile stuff

.travis.yml
```diff
@@ -38,6 +48,9 @@ script:
     - popd
     - sleep 3
     - DISPLAY=:99 python -m unittest discover -s selenium_tests -t . -v
+    - pushd webclient
```
stefanoborini: same

webclient/.bowerrc
```diff
@@ -0,0 +1,4 @@
+{
+  "directory": "bower_components",
```
stefanoborini: We don't put power stuff in bower_components. It should be components, and we are already doing so, so it's probably in the bower.json

webclient/package.json
```diff
@@ -0,0 +1,44 @@
+{
```
stefanoborini: check if there's a way of having the root package.json call this one in the npm documentation.

webclient/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "name": "webclient",
+  "version": "0.0.0",
+  "description": "Small description for webclient goes here",
```
stefanoborini: description

webclient/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "webclient",
+  "version": "0.0.0",
```
stefanoborini: version. set it equal to the overall version of our software

webclient/package.json
martinRenou: To call the package.json ?

stefanoborini: to refer to, so that you can build both at once.

webclient/.bowerrc
martinRenou: I think the directory is named bower_components by default in order to prevent confusions between bower dependencies and Ember components. But I can rename it components if you'd prefer.

martinRenou: And I think it has to stay here instead of being in remoteappmanager/static/components because it will be packaged and minified in the final javascript bundle by "ember build", I may be wrong

webclient/package.json
martinRenou: I can't found anything about that, but we could create a script in the first package.json that could run "npm install; cd webclient; npm install; cd ..".
But if we do that, it means that instead of running "npm install" as usual to install the dependencies, we will run "npm run install", and I'm not sure if it's a good practice.

stefanoborini:
Going with vue

PR #399, opened 2017-04-13
Switch to react, first step

martinRenou:
- Installation of react
- Setup build process
- Add first react component to the html page


stefanoborini:
Ok, let go with vue.

PR #401, opened 2017-04-18
Remove everything in order to start from scratch

martinRenou:


stefanoborini:
Just change ApplicationListView. That will also give us a measure of how smaller the codebase becomes.

PR #402, opened 2017-04-18
Replace application list view with a Vue ViewModel [WIP]

martinRenou:


martinRenou:
That's still in progress.
What I have:
- The application list correctly displayed
What I don't have:
- The ViewModel is not correctly synchronized with the model. The model should be an attribute of the ViewModel, so that the ViewModel knows when the model changed.
- There's no stop button

bower.json
```diff
@@ -1,32 +1,33 @@
 {
-    "name": "simphony-remote-deps",
```
stefanoborini: Use four spaces for indentation. I checked the preferred style guide, and most of the opinions are for four spaces, like in python, including Crockford. We don't have to save space, minification will do that (if ever, js files are cached anyway)

stefanoborini: Ok, correction. Google uses two. uhm....

stefanoborini: ok, keep two.

stefanoborini: Better, keep four for now, and we'll switch to two after this PR is in

remoteappmanager/templates/home.html
```diff
@@ -17,11 +17,51 @@
...
+  <ul id="applist" class="sidebar-menu">
     <li class="header">APPLICATIONS</li>
-  </ul>
 
-  <ul class="sidebar-menu" id="applist">
+    <!--
+      We initialize this element with "display: none" to prevent displaying
+      "No applications found" and the loading spinner at the same time
+    -->
+    <li v-show="!loading && model.app_list.length === 0" style="display: none;">
```
martinRenou: There's no gain to use `v-if` and `v-else` instead of two `v-show` in terms of lines of code. And my problem is that it changes the behavior of the dom elements, `v-if`will result in rendering the element only if the statement is true, `v-show`only changes the style attribute `display`of the element, so I have to change the selenium tests which relies on the style attribute

PR #403, opened 2017-04-21
Update against new tornado webapi

stefanoborini:


remoteappmanager/static/js/home/models.js
```diff
@@ -25,54 +25,15 @@ define([
...
-                for (i = 0; i < ids.length; i++) {
-                    var id = ids[i];
-                    resources.Application.retrieve(id)
-                        .done(done_callback(i))
-                        .fail(fail_callback(i));
-
+            .done(function (identifiers, items) {
+                var result = [];
+                for (var key in items) {
+                    if (!items.hasOwnProperty(key)) {
```
martinRenou: I think you can replace this loop with the if statement by something like:
```javascript
Object.keys(items).forEach(function(key){
    result.push(items[key]);
});
```

PR #407, opened 2017-04-27
Refactor model

martinRenou:
- [x] The loading property now belongs to the model
- [x] The application list is sorted by application name
- [x] The views are correctly instantiated with the right model instead of a "default" model
- [x] The start_application and stop_application methods are in the model

### DIFF
--- a/remoteappmanager/static/js/home/controller.js
+++ b/remoteappmanager/static/js/home/controller.js
@@ -15,11 +15,15 @@ require([
     // It is only synchronized at initial load.
     var model = new models.ApplicationListModel();
 
-    var app_list_view = new application_list_view.ApplicationListView();
-    var app_view = new application_view.ApplicationView();
-
-    $.when(model.update()).done(function () {
-        app_list_view.model = model;
-        app_view.model = model;
+    var app_list_view = new application_list_view.ApplicationListView({
+        el: '#applist',
+        data: function() { return { model: model }; }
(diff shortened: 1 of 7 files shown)


remoteappmanager/static/js/home/controller.js
```diff
@@ -15,13 +13,15 @@ require([
...
-        app_list_view.loading = false;
+    new application_list_view.ApplicationListView({ // jshint ignore:line
+        el: '#applist',
+        data: function() { return { model: model }; }
+    });
 
-        app_list_view.model = model;
-        app_view.model = model;
+    new application_view.ApplicationView({ // jshint ignore:line
+        el: 'div.content-wrapper',
```
martinRenou: add an ID to that specific snippet.

remoteappmanager/static/js/home/views/application_view.js
```diff
@@ -2,26 +2,11 @@ define([
     'jquery',
     "urlutils",
     "utils",
-    "gamodule",
-    "dialogs",
-    "home/models",
-    '../../components/vue/dist/vue.min',
-    "jsapi/v1/resources"
-], function ($, urlutils, utils, gamodule, dialogs, models, Vue, resources) {
+    '../../components/vue/dist/vue.min'
```
martinRenou: you can use vue instead of vue.min

PR #411, opened 2017-04-28
Create google analytics observer

martinRenou:


remoteappmanager/static/js/gamodule.js
```diff
@@ -7,46 +7,31 @@ define([
...
+    GaObserver.prototype.trigger_application_starting = function(name) {
+        this.ga("send", "event", {
+            eventCategory: "Application",
+            eventAction: "start",
+            eventLabel: name
+        });
+    };
 
     return {
         init: init, // For testing purpose
```
stefanoborini: at this point, init could be a method of GaObserver. Keep it separate, because we still need to patch it at the test level

PR #414, opened 2017-05-03
Switch to vue (user frontend)

stefanoborini:
Switch the user frontend to Vue.


stefanoborini:
Before merging, we need to do a full pass to rename all entities to the vue standard naming e.g. selected_index -> selectedIndex

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,57 @@
...
-    var ns = {
-        from_tag: from_tag
+        return { resolution: resolution };
     };
 
-    return $.extend(ns, configurables);
-    
+    // Export all your configurable models here
+    return {
+        resolution: ResolutionModel,
```
stefanoborini: tag: { model: theModel, component: theComponent}

remoteappmanager/static/js/home/models.js
```diff
@@ -41,95 +41,166 @@ define([
...
-                self._update_configurables(data_idx);
-                self._update_status(data_idx);
-            }
-        });
+            appData.forEach(function(applicationData) {
+                var app = {
+                    appData: applicationData,
+                    // Default values, will be overwritten
+                    status: Status.STOPPED,
+                    delayed: true,
```
stefanoborini: document what delayed is.

PR #417, opened 2017-05-03
Fix search form behavior

martinRenou:
Deploys correct behavior for search, restricting to the applications that are relevant to the search itself.

### DIFF
--- a/remoteappmanager/static/js/home/views/application_list_view.js
+++ b/remoteappmanager/static/js/home/views/application_list_view.js
@@ -30,9 +30,8 @@ define([
               '  </li>' +
 
               '  <!-- Application list -->' +
-              '  <li v-for="(app, index) in model.app_list"' +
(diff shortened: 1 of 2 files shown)


remoteappmanager/static/js/home/views/application_list_view.js
```diff
@@ -51,14 +45,29 @@ define([
...
             '</section>' +
-            '<!-- /.sidebar -->'
+            '<!-- /.sidebar -->',
+
+        data: function() {
+            return { 'search_input': '' };
+        },
+
+        computed: {
+            'entries_visible': function() {
```
stefanoborini: you don't need a visibility array. What you need is a filtering operation over the model vs represented model.

stefanoborini: Also, this one is new behavior so you need a selenium test for it

PR #420, opened 2017-05-04
Resolution configurable is now a Vue component

martinRenou:


martinRenou:
I am not really happy with this PR. I would like to reduce the amount of code for the component.

jstests/tests/home/test_configurables.js
```diff
@@ -1,26 +1,17 @@
...
+        var resolution = $.extend(true, {}, configurables["resolution"]);
+
         assert.equal(resolution.tag, "resolution");
-        assert.equal(resolution.resolution, "Window");
-        
-        resolution.resolution = "1024x768";
-        assert.equal(resolution.as_config_dict().resolution, "1024x768");
-    });
-    
-    QUnit.test("view", function (assert) {
```
stefanoborini: in principle you can still use the view on a configurable component.

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,53 @@
...
+            '  <select class="form-control" v-model="sel_value">' +
+            '    <option v-for="resolution_option in resolution_options">' +
+            '      {{resolution_option}}' +
+            '    </option>' +
+            '  </select>' +
+            '</div>',
+
+        data: function() {
+            return {
+                sel_value: this.value,
```
stefanoborini: selected_value

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,53 @@
...
-        // parameter under the key given by the tag member. 
-        var resolution = this.resolution;
-        
-        if (resolution === 'Window') {
-            var max_size = utils.max_iframe_size();
-            resolution = max_size[0]+"x"+max_size[1];
+        template:
+            '<div class="form-group">' +
+            '  <label>Resolution</label>' +
+            '  <select class="form-control" v-model="sel_value">' +
```
stefanoborini: call it selected_value

remoteappmanager/static/js/home/views/application_view.js
```diff
@@ -42,7 +42,15 @@ define([
             '          </ul>' +
 
             '          <h4>Configuration</h4>' +
-            '          <form class="configuration"><fieldset></fieldset></form>' +
+            '          <form class="configuration">' +
+            '            <fieldset v-if="current_app.configurables.length === 0">No configurable options for this image</fieldset>' +
+            '            <fieldset v-else :disabled="current_app.is_starting()">' +
+            '              <component v-for="configurable in current_app.configurables"' +
+            '                         :is="configurable.tag"' +
+            '                         :value="configurable.value"' +
```
stefanoborini: sync

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,53 @@
...
-    var view_template = hb.compile(
-        '<div class="form-group">' +
-        '<label for="resolution-model-{{index}}">Resolution</label>' +
-        '<select class="form-control" id="resolution-model-{{ index }}">' +
-        '{{#each options}}' +
-           '<option value="{{this}}">{{this}}</option>' +
-        '{{/each}}' +
-        '</div>'
-    );
+    var resolution_tag = 'resolution';
```
stefanoborini: Vue recommends to use a hyphen.

stefanoborini: maybe resolution-component?

stefanoborini: make sure that you use reasonable differentiation between the configurable tag (e.g. "resolution") and the component tag (which is e.g. "resolution-component")

remoteappmanager/static/js/home/models.js
```diff
@@ -121,15 +120,15 @@ define([
...
             // If this returns null, the tag has not been recognized
             // by the client. skip it and let the server deal with the
             // missing data, either by using a default or throwing
             // an error.
-            var ConfigurableCls = configurables.from_tag(tag);
+            var configurable = configurables[conf_name];
 
-            if (ConfigurableCls !== null) {
-                app.configurables.push(new ConfigurableCls());
+            if (configurable !== null) {
```
stefanoborini: I think it's undefined, not null, if you have an element that is not there.

remoteappmanager/static/js/home/views/application_view.js
martinRenou: The `.sync`doesn't work like expected... The child `selected_value` is not correctly synchronized with the parent value.

jstests/tests/home/test_configurables.js
```diff
@@ -1,26 +1,17 @@
 define([
+    "jquery",
     "home/configurables"
-], function (configurables) {
+], function ($, configurables) {
```
martinRenou: Here I use jQuery for the `extend` method which does a deep copy of the dictionary representing the model of the configurable. Underscore provides an `extend`method too but it doesn't do a deep copy of the dictionary.
Instead of returning a list of dictionaries representing the models of configurables in `configurable.js`I could return a list of classes. So we get rid of this deep copy of dictionaries.

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,56 @@
...
-        };
-    };
-   
-    // Define all your configurables here.
-    var configurables = {
-        ResolutionModel: ResolutionModel
+    });
+
+    // Your configurable class must implement tag and value attributes and as_config_dict method
+    var resolution_model = function() {
```
stefanoborini: Should be ResolutionModel (as a class)

stefanoborini: which brings me to: do we really need it? I mean, technically, we could
- have only the component.
- have a component method `as_config_dict` that extracts the data from the component.
- have the parent call this method when submit is called, aggregate all responses, and send them as part of the request

remoteappmanager/static/js/home/views/application_view.js
```diff
@@ -42,7 +42,15 @@ define([
             '          </ul>' +
 
             '          <h4>Configuration</h4>' +
-            '          <form class="configuration"><fieldset></fieldset></form>' +
+            '          <form class="configuration">' +
+            '            <fieldset v-if="current_app.configurables.length === 0">No configurable options for this image</fieldset>' +
+            '            <fieldset v-else :disabled="current_app.is_starting()">' +
+            '              <component v-for="configurable in current_app.configurables"' +
+            '                         :is="configurable.tag + \'-component\'"' +
+            '                         :value="configurable.value"' +
```
stefanoborini: you can use .sync in this case. it should work, we are using vue 2.3

remoteappmanager/static/js/home/configurables.js
martinRenou: I totally agree with this, that's what I wanted to do first. But I faced some problems with this implementation:
- How to call the `as_config_dict` method from the parent ? The parent here is `ApplicationView` and we won't have access to the child component and this method in the Javascript code, as far as I understand.
-  How to define the value of resolution of each application without a `model` object ?

remoteappmanager/static/js/home/views/application_view.js
martinRenou: As I said in the other comment, I tried to use `.sync` but it didn't work. `selected_value` wasn't correctly synchronized with the parent's value. I'm not sure what is going on.

remoteappmanager/static/js/home/configurables.js
stefanoborini: - https://vuejs.org/v2/guide/components.html#Child-Component-Refs
- each component uses a different data object. the parent does not need to know anything about the types of the children. It just needs to know that, once done, it will go through the list and invoke `as_config_dict`

martinRenou: Ok, great thank you :) So with this I can define `as_config_dict` directly in the component.
We still have the issue of defining a value associated to each application, do you have any idea ?

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,59 @@
...
-    };
-    
-    var from_tag = function (tag) {
-        // Given a tag, lookup the appropriate configurable and
-        // return it. If the tag matches no configurable, returns null
-        for (var conf in configurables) {
-            if (configurables[conf].prototype.tag === tag) {
-                return configurables[conf];
+
+                return { resolution: { resolution: resolution } };
```
martinRenou: This config dict is really weird, maybe we should change the representation needed by the container ?

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,59 @@
...
-    };
+    });
 
-    var ns = {
-        from_tag: from_tag
+    // Your configurable class must implement tag and value attributes
+    var ResolutionModel = function() {
+        this.tag = resolution_conf_tag;
+        this.value = 'Window';
+        this.component = resolution_component;
```
martinRenou: This line is only for the tests. It's really weird to have the component constructor in the model. I should change this.

martinRenou: By the way, I couldn't find a proper way to set the value for each application without creating a model.

remoteappmanager/static/js/home/configurables.js
stefanoborini: why weird?

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,59 @@
...
-        '<select class="form-control" id="resolution-model-{{ index }}">' +
-        '{{#each options}}' +
-           '<option value="{{this}}">{{this}}</option>' +
-        '{{/each}}' +
-        '</div>'
-    );
+    var resolution_conf_tag = 'resolution';
+    var resolution_component = Vue.component(resolution_conf_tag + '-component', {
+        // Your configurable must have a "value" property
+        props: ['value'],
```
stefanoborini: data?

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,59 @@
...
+            '  <select class="form-control" v-model="selected_value">' +
+            '    <option v-for="resolution_option in resolution_options">' +
+            '      {{resolution_option}}' +
+            '    </option>' +
+            '  </select>' +
+            '</div>',
+
+        data: function() {
+            return {
+                selected_value: this.value,
```
stefanoborini: data that you accept is a bundle that contains specific information for the internals of the ui.
so in this case you have to extract data["resolution"] and put it here as the value for this specific widget of the UI.

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,59 @@
...
-        return null; 
-    };
+    });
 
-    var ns = {
-        from_tag: from_tag
+    // Your configurable class must implement tag and value attributes
+    var ResolutionModel = function() {
+        this.tag = resolution_conf_tag;
+        this.value = 'Window';
```
stefanoborini: this.data = {"resolution": "Window"}

remoteappmanager/static/js/home/configurables.js
stefanoborini: just return for consistency with data that you put in.

remoteappmanager/static/js/home/configurables.js
stefanoborini: config_dict ?

remoteappmanager/static/js/home/configurables.js
```diff
@@ -1,85 +1,59 @@
...
+            };
+        },
+
+        watch: {
+            value: function() {this.selected_value = this.value;}, // model -> view update
+            selected_value: function() {this.$emit('update:value', this.selected_value);} // view -> model update
+        },
+
+        methods: {
+            as_config_dict: function() {
```
stefanoborini: this should be in the model. All your interaction should be with the model, and the component is just a view that is inserted in the right place.

remoteappmanager/static/js/home/models.js
```diff
@@ -121,15 +121,15 @@ define([
         // to its client-side model.
         app.configurables = [];
 
-        app.app_data.image.configurables.forEach(function(tag) {
+        app.app_data.image.configurables.forEach(function(conf_name) {
```
stefanoborini: should be tag

PR #422, opened 2017-05-04
Remove the use of Jquery in Vue objects

martinRenou:


remoteappmanager/static/js/home/controller.js
```diff
@@ -51,5 +51,7 @@ require([
         ga_observer.trigger_application_starting(application.app_data.image.name);
     });
 
+    app_list_view.$on('focus_iframe', function() { app_view.focus_iframe(); });
```
stefanoborini: I think you can just check on click event instead.

martinRenou: I tried, it didn't work. So I just renamed the event to `entry_clicked`

remoteappmanager/static/js/home/controller.js
```diff
@@ -51,5 +51,7 @@ require([
         ga_observer.trigger_application_starting(application.app_data.image.name);
     });
 
+    app_list_view.$on('entry_clicked', function() { app_view.focus_iframe(); });
```
stefanoborini: only clicked. You don't need another event. clicked is already an event.

martinRenou: I can rename it clicked, but I will still need to emit the event with the `$emit` method. You want me to rename it ?

martinRenou: As I said in the other comment, it sounds like "clicked" or "click" are not built-in events for the `$on` method of Vue. I think it's not really related to the javascript method `addEventListener("click", myFunction)`. We can only use the `$on`method with custom events and trigger them with the `$emit` method.

martinRenou: I also tried `app_list_view.$on('click.native', myFunction)` according to the [documentation](https://vuejs.org/v2/guide/components.html#Using-v-on-with-Custom-Events) but it sounds like the `.native` modifier doesn't work for `$on`, but only for `v-on` in templates.

martinRenou:
@stefanoborini Ok to merge or you would prefer to rename the event from `entry_clicked` to `clicked`?

PR #423, opened 2017-05-05
Add transition effect

martinRenou:
Add a fade effect. This effect will be applied only when starting and stopping the application, not when switching from a running application to a stopped application.


martinRenou:
We agreed that it should be a simple fade instead of a sliding fade

selenium_tests/test_hide_application.py
```diff
@@ -1,23 +1,27 @@
...
             search_box = driver.find_element_by_name("q")
             search_box.clear()
             search_box.send_keys('foobarheho')
 
-            self.wait_for(
-                lambda:
-                driver.find_element_by_css_selector(
-                    "#applistentries > li").value_of_css_property(
-                        'display') == "none")
+            WebDriverWait(driver, 3).until(
```
stefanoborini: you can remove this.

PR #425, opened 2017-05-05
Server side admin

stefanoborini:
Changes to the server side admin webapi to support switch to Vue based interface.

### DIFF
--- a/remoteappmanager/webapi/admin/application.py
+++ b/remoteappmanager/webapi/admin/application.py
@@ -50,6 +50,13 @@ class ApplicationHandler(ResourceHandler):
     @gen.coroutine
     @authenticated
     def items(self, items_response, **kwargs):
+        """Produces a list of Application items in the items_response object.
+
+        Parameters
+        ----------
+        items_response: ItemsResponse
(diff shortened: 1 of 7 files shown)


remoteappmanager/webapi/admin/application.py
```diff
@@ -46,3 +46,16 @@ def create(self, resource, **kwargs):
             raise exceptions.Unable()
 
         resource.identifier = str(id)
+
+    @gen.coroutine
+    @authenticated
+    def items(self, items_response, **kwargs):
+        db = self.application.db
```
martinRenou: You can maybe add a docstring

remoteappmanager/webapi/admin/user.py
```diff
@@ -45,3 +45,12 @@ def create(self, resource, **kwargs):
             raise exceptions.Exists()
         except db_exceptions.UnsupportedOperation:
             raise exceptions.Unable()
+
+    @gen.coroutine
+    @authenticated
+    def items(self, items_response, **kwargs):
+        users = self.application.db.list_users()
```
martinRenou: Same here

PR #427, opened 2017-05-08
Fixes issue with indexing after search filter.

### LINKED ISSUE #426: model.selected_index doesn't correspond to the active element in the list
In the branch [switch_to_vue](/simphony/simphony-remote/tree/switch_to_vue) we included a bug with the PR #417.

When we type something in the search box, the list of visible entries changes. So the selected_index in the model (which is linked to the original list) doesn’t correspond to the same index in the new `visible_list` computed property.

So the problem is that the active element in the list doesn’t correspond to the current selected application.

martinRenou:
Fixes #426


remoteappmanager/static/js/home/views/application_list_view.js
```diff
@@ -65,6 +65,12 @@ define([
                     return app_name.indexOf(this.search_input.toLowerCase()) !== -1;
                 }.bind(this));
             }
+        },
+
+        methods: {
+            index_of: function(app) {
```
stefanoborini: indexOf

PR #431, opened 2017-05-09
Now use dev mode with Vue

martinRenou:


martinRenou:
Ok, I would expect that loading the minified version or the dev version wouldn't affect the tests. However, I suspect that jQuery is the reason of the failure because an error `$ is not a function` is raised in the console with the dev version (hidden with the min version). So once the PR #422 is merged I think the tests will pass.

PR #433, opened 2017-05-09
Export Vue filters in a seperate file

martinRenou:
Moved vue filters in a separate file under static/js/vue. This directory will also contain the generic components and form our toolkit.


remoteappmanager/static/js/filters.js
```diff
@@ -0,0 +1,22 @@
+define([
```
stefanoborini: we need a standard place for vue stuff at this point. I have the same problems with components.
I would say create a directory /static/js/vue/. Inside, put filters.js. I will also do the same, and have a subdir components, where I'll put the initial toolkit.

PR #440, opened 2017-05-10
Introduced data-table and confirm-dialog components: Containers

stefanoborini:
Uses new components in the Containers table.


remoteappmanager/static/js/admin/vue-components/ContainersView.js
```diff
@@ -1,96 +1,99 @@
...
-          this.updateTable();
+        stopAction: function(row) {
+          this.stopContainerDialog.containerToStop = row[0];
+          this.stopContainerDialog.show = true;
         },
-        stopAction: function(index) {
-          this.containerToStop = this.containers[index].identifier;
-          this.showStopContainerDialog = true;
+        stopContainer: function () {
+          if (this.stopContainerDialog.containerToStop === null) {
```
martinRenou: Can this happen ?

stefanoborini: not in the current code, but if someone decides to call it programmatically and does not fill containerToStop it will just do nothing, although probably it's better if we let the error happen.

PR #446, opened 2017-05-11
Change naming convention from snake_case to camelCase

martinRenou:


remoteappmanager/static/js/vue/filters.js
```diff
@@ -1,20 +1,20 @@
...
-], function (Vue, urlutils) {
+], function (Vue, urlUtils) {
     'use strict';
 
-    Vue.filter('icon_src', function(icon_data) {
+    Vue.filter('iconSrc', function(icon_data) {
         return (
             icon_data ?
             'data:image/png;base64,' + icon_data :
-            urlutils.path_join(
```
stefanoborini: pathJoin.

remoteappmanager/static/js/vue/filters.js
```diff
@@ -1,20 +1,20 @@
 define([
     "components/vue/dist/vue",
     "urlutils"
-], function (Vue, urlutils) {
+], function (Vue, urlUtils) {
```
stefanoborini: question is if we should use the same for modules, or just stick to the filename (all lowercase).

But we use Vue... but then again, Vue is also weird because it's a class...

martinRenou: I just wanted to stay consistent in all files (except for the variables which come from the server). But if I rename it `urlUtils` I should maybe rename the file as well ?

stefanoborini: google style guide says lowercase for filenames.. so no, the file is urlutils...

Leave it as is.

PR #458, opened 2017-05-17
Create Vue files and use vue-loader plugin of Webpack

martinRenou:
The changes are:
- [x] configure Webpack so that we can load `.vue` files
- [x] transpose old `.js` files containing Vue components into `.vue` files. It only results in splitting the template and the javascript into two separate tag elements `<template>` and `<script>` for each component.
- [x] application_list_view.js -> ApplicationListView.vue
- [x] user/views -> user/vue-components (to stay consistent with the admin)
- [x] configurables.js -> vue-components/configurables/Resolution.vue
- [x] user/models -> user/ApplicationListModel.js


martinRenou:
I think it's enough for this PR. I will put the CSS in Vue files in a next PR.

PR #459, opened 2017-05-18
Move the CSS in each corresponding component

martinRenou:


remoteappmanager/static/js/user/vue-components/ApplicationListView.vue
```diff
@@ -100,3 +100,157 @@
     }
   });
 </script>
+
+<style>
```
stefanoborini: try to scope it

PR #462, opened 2017-05-22
Create a component for the topbar

### LINKED ISSUE #376: Make stop icon more visible

### LINKED ISSUE #415: Put application name and close button in the top bar
We should have the application name and the close button in the top bar, just after the "Simphony Remote" name.

martinRenou:
This component (named ApplicationLabel) is a big dropdown button showing the current application running and its icon. When clicking on this button it displays application settings containing only the stop button for now. More settings could come later.

The stop buttons has been removed from the application list view.

This PR comes with big changes on the style of the user page. The top bar is now white, application icons are now squares with border radius of 20% instead of circles, the dropdown of the resolution widget is now less large.

Closes #415 
Closes #376


frontend/user/vue-components/ApplicationListView.vue
```diff
@@ -1,15 +1,5 @@
 <template>
   <section class="sidebar">
-  <!-- Error dialog -->
```
stefanoborini: How do we report this now? I can't remember what we settled on.

martinRenou: Now it's in ApplicationLabel.vue, and I had some test for it in the corresponding test file

martinRenou: Fixed #467

PR #465, opened 2017-05-22
Fix issue when ApplicationView is not rendered because of an empty app list

### LINKED ISSUE #461: Error when the application list is empty
When the application list is empty, an error occurred "this.$el is undefined" in `ApplicationView.vue` because the component is simply not rendered.

martinRenou:
Fixes #461

### DIFF
--- a/frontend/user/vue-components/ApplicationView.vue
+++ b/frontend/user/vue-components/ApplicationView.vue
@@ -135,9 +135,9 @@
         return utils.maxIframeSize();
       },
       focusIframe: function() {
-        let iframe = this.$el.querySelector('iframe');
-        if(iframe !== null) {
-          iframe.focus();
+        if(this.$el.querySelector !== undefined && // In case $el is not rendered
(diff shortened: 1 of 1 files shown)


frontend/user/vue-components/ApplicationView.vue
```diff
@@ -135,9 +135,9 @@
         return utils.maxIframeSize();
       },
       focusIframe: function() {
-        let iframe = this.$el.querySelector('iframe');
-        if(iframe !== null) {
-          iframe.focus();
+        if(this.$el.querySelector !== undefined && // In case $el is not rendered
```
stefanoborini: I think you are not using the right approach. if the element is not rendered, you should check if the element is not rendered. You are probing the existence of a routine that happens to be correlated to the existence of the element, but it's not the thing that is the core issue.

martinRenou: You are absolutely right

PR #481, opened 2017-05-31
Fix close button issue

### LINKED ISSUE #479: Stop application can be clicked and triggers red dot status indicator
If you click on the "stop application" entry in the menu, even if it's disabled, it still triggers something, because the red dot indicator starts flashing as if the application is being turned off.

martinRenou:
As it's not a "button" element but a "li" element, there is no disable attribute, we only play with the disabled class which change the visual aspect but doesn't prevent the click event to be triggered.

I fixed this by changing the `disabled` class which now prevents the mouse events. 

Closes #479

### DIFF
--- a/frontend/tests/tests/user/test_application_label.js
+++ b/frontend/tests/tests/user/test_application_label.js
@@ -1,3 +1,4 @@
+let $ = require("jquery");
 let Vue = require("vuejs");
 let ApplicationListModel = require("user/ApplicationListModel");
 let ApplicationLabel = require("user/vue-components/ApplicationLabel");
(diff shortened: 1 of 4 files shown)


stefanoborini:
The solution is good, but I was wondering if we can actually prevent the click. Just curious.

martinRenou:
I changed it, it now prevents the click

stefanoborini:
No, leave the check. It makes sense it's there. You can't operate if it's not running, but it should also prevent the click.

martinRenou:
Ok, so I should do the same for "StartApplication" and add a check in order to stay consistent

stefanoborini:
Please add a test that the thing is actually not starting or stopping when those conditions happen.

PR #482, opened 2017-05-31
Fix tables display issues

martinRenou:
We now use adminlte tables which are more responsive


stefanoborini:
Leave the button as it was before and the generalActions as well. We'll find a way of embedding the button in another component at a later stage.

martinRenou:
As you asked, I left the button as it was before. Excepted that I played with the CSS to display it on top of the datatable instead of on its side. It involves flex-box, but it will be susceptible to be removed when putting the button in the adminlte box header.

PR #506, opened 2017-06-05
It is now more obvious when buttons are disabled

### LINKED ISSUE #487: Deselected entries are not easy to see
We already discussed about this. The deselected entries are really hard to see. I wonder if we can do something to the CSS to solve the problem. Just as a reminder.

martinRenou:
ApplicationLabel changes.
- This is now a notification-menu of Admin-lte, which looks better than a simple dropdown
- Opacity changes when the stop button or the share button are not enabled.

Closes #487


frontend/user/vue-components/ApplicationLabel.vue
```diff
@@ -1,42 +1,44 @@
...
         <span>{{ currentApp.appData.image | appName }}</span>
       </a>
 
-      <ul class="dropdown-menu" role="menu">
-        <li class="disabled">
-          <a href="#">Application settings</a>
-        </li>
-        <li role="separator" class="divider"></li>
+      <ul class="dropdown-menu">
+        <li class="header">Application settings</li>
```
stefanoborini: I would remove this entry here. It conflicts with HIG and brings in a problem we should not have to begin with.

frontend/user/vue-components/ApplicationLabel.vue
```diff
@@ -1,42 +1,44 @@
 <template>
   <ul v-if="currentApp !== null" class="nav navbar-nav">
-    <li class="dropdown">
+    <li class="dropdown notifications-menu">
```
stefanoborini: removing the menu "section title" means that you don't need notification menu anymore.

Issue #513, opened 2017-06-06
Rename and relocate "share" option


stefanoborini:
Rename the share option as "Share session" (lacking better choice) and move it above the quit entry.

martinRenou:
Done #519

Issue #512, opened 2017-06-06
Stop application should be renamed as "Quit"


stefanoborini:
The "stop application" menu entry should be "quit" instead. The reason is that it might deceive in thinking that the application is only suspended, while in reality it just quits and deletes everything.

martinRenou:
Done #519

Issue #508, opened 2017-06-06
Modal dialogs for confirm and share.


martinRenou:
Introduce dialogs for stopping the application and sharing.

- The dialog for stopping will ask for confirmation to stop.
- The dialog for share will give the copy and paste URL, and eventually will also be used to add which users are allowed to see the desktop.

martinRenou:
Done #519

PR #519, opened 2017-06-08
Application label refactor

### LINKED ISSUE #493: Share button improvements
The share button should behave like the "Clone" button in Github and display the url of the application which is saved in the clipboard, so that the user can still share the application url even if his browser doesn't support the save to clipboard feature

martinRenou:
- Rename "Stop application" link to "Quit"
- Rename "Share (copy url)" link to "Share session"
- Move share link above the quit link
- Those links now opens modal dialogs
  - The share link opens a dialog with a a share button (which looks like the github clone button)
  - The quit link opens a confirm dialog before stopping the application

Closes #493 #512 #508 #513


frontend/user/vue-components/ApplicationLabel.vue
```diff
@@ -52,13 +78,11 @@
 
   module.exports = Vue.extend({
     data: function() {
-      return { clipboardSupported: Clipboard.isSupported() };
+      return { shareDialogVisible: false, quitDialogVisible: false };
```
stefanoborini: shareDialog.visible

PR #521, opened 2017-06-13
Remove Jinja and use Tornado templates

### LINKED ISSUE #474: Remove jinja?
With the backend mostly providing json data, and the main pages now being way simpler, I wonder if we can get rid of jinja and just rely on the basic tornado templating mechanism. It's very similar to jinja, except for being not as powerful, so I suspect the transition will be almost painless, given that we use quite basic features now.

martinRenou:
Closes #474


remoteappmanager/handlers/admin/admin_home_handler.py
```diff
@@ -9,4 +9,4 @@ class AdminHomeHandler(BaseHandler):
     @web.authenticated
     @gen.coroutine
     def get(self):
-        self.render('admin/page.html')
```
martinRenou: I changed the name of the admin and user page files because when doing `{% extends "page.html" %}` they were loading themselves resulting on an infinite loop.

remoteappmanager/templates/admin_page.html
```diff
@@ -1,6 +1,6 @@
 {% extends "page.html" %}
 
-{% set skin = "skin-red" %}
```
martinRenou: With tornado `{% set skin = ... %}` only sets a local variable which is not available in the parent scope. Unfortunately, according to https://github.com/tornadoweb/tornado/pull/1323, an equivalent implementation in Tornado of the behavior of Jinja is not available yet. A workaround is to use a block.

remoteappmanager/templates/error.html
```diff
@@ -1,28 +1,23 @@
...
+      {{status_code}} : {{status_message}}
+    </h1>
+  {% end %}
   {% block error_detail %}
-  {% if message %}
-  <p>
-    {{message}}
-  </p>
-  {% endif %}
-  {% if message_html %}
```
martinRenou: I couldn't find any other reference to `message_html`, maybe an old unused anymore variable ?

remoteappmanager/templates/page.html
```diff
@@ -95,36 +93,35 @@
...
-      {% endblock %}
+      {% end %}
     </div>
   </div>
 
   <div id="error-dialog-container"></div>
 
   <script src="{{ static_url("dist/sharedDependencies.js") }}"></script>
 
+  <script type="text/javascript">
```
martinRenou: In tornado we don't have the possibility to call `super ()` in blocks, moving this script out of the block `script_init` was the only solution that I found.

remoteappmanager/templates/page.html
```diff
@@ -52,79 +50,80 @@
...
+          tracking_id: "{{ analytics['tracking_id'] }}"
+        },
+      {% end %}
+    }
+  </script>
   {% block script_init %}
-    <script type="text/javascript">
-      window.apidata = {
-        base_url: "{{base_url}}",
-        prefix: "{{prefix}}",
```
martinRenou: Same here, I couldn't find any other reference to prefix

martinRenou: Jinja wasn't complaining but tornado is less permissive. With tornado code in the brackets is python code so it raises a `name "prefix" is not defined`

stefanoborini:
Checked it once. I am unsure about a few things. Leaving for later.

remoteappmanager/templates/page.html
```diff
@@ -52,79 +50,80 @@
...
-                    {% block login_widget %}
-                    <a id="logout" href="{{ logout_url }}" class="btn btn-default btn-flat"><i class="fa fa-sign-out"> Sign out</i></a>
-                    {% endblock %}
-                  </div>
-                </li>
-              </ul>
-            </li>
-          </ul>
-        </div>
+        {% if user is not None %}
```
stefanoborini: When do we have user == None? (maybe we already talked about it, but I don't remember)

martinRenou: I had it when rendering the `error.html` template

stefanoborini: ok, that's suspicious... why?

martinRenou: You're right, I shouldn't. So maybe I'm wrong, maybe I had an other reason but I can't remember..

stefanoborini: Ok, I suspect that it's because if you login and another user connects to the server, user is not defined (you are not passing the right credentials for the process) and in that case you issue an error message. but the error message cannot be shown because it depends on a template that requires the user.

So the best solution is to have an error message with a different base template.

martinRenou: Should we do that in this PR ?

stefanoborini: Yes please, it makes sense that we rationalise our templates. They are going away anyway, but as a stepstone it's probably the best strategy

Issue #504, opened 2017-06-05
Show images that are present in the docker but not yet present in the database: image availability


stefanoborini:
If we have some images in docker, we should display them in the application list of the admin anyway, and mark them as disabled. If an image is not in the database it should be considered as such.

If the image is in the database, it will have an additional boolean column for when it's enabled.
Only enabled images will be runnable by the users.

Only images that are fully supported will be shown. You can use the application type label to distinguish which ones are valid simphony-remote images.

stefanoborini:
you have four classes of images:

1. images that are NOT part of simphony-remote
2. images that are part of simphony-remote but are NOT in the db.
3. images that are part of simphony-remote but are in the db, and are disabled
4. images that are part of simphony-remote but are in the db, and are enabled.

You don't want to show 1.
you want to show 2, but mark them as disabled.
you want to show 3 and 4, and are enabled or disabled accordingly.

Distinguish 1 via the application type.

martinRenou:
In the new application dialog it would be great to have a combo-box with available images on docker instead of typing the image name. This means having a backend which supports docker related requests.

PR #525, opened 2017-06-19
Show images that are present in the docker but not yet present in the database: image availability [WIP]

martinRenou:
- [x] Show images that are present in the docker but not in the database
- [x] Add the possibility to add those images to the database
- [x] Remove datatable's global action for the application view
- [ ] Improve the style of the row actions by using a checkbox instead of buttons "Add" and "Remove"

Partially fix issue #504


martinRenou:
We would like to have an "enabled" flag for images in the database. The admin user would see in the interface the list of docker images related to simphony remote (even if they are not in the database). Images that are not in the database should have the enabled flag set to false. The admin could enable or disable those images, if he sets to enabled an image which is not in the database, this image would be added to the database with the enabled flag set to true.

This requires too much changes and a database migration, we will think about that before opening an other PR

PR #527, opened 2017-06-21
Fix error when stopping a container from the admin interface

### LINKED ISSUE #526: Error when stopping a container from the admin interface
When stopping a container from the admin interface the modal dialog is closed after the application is stopped. This results in an error when double clicking on the "Ok" button of the modal dialog, because it will try to stop twice and an error message will appear. 
We can fix this issue by closing the error dialog before trying to stop the application.

martinRenou:
Closes #526

### DIFF
--- a/frontend/admin/vue-components/ContainersView.vue
+++ b/frontend/admin/vue-components/ContainersView.vue
@@ -73,13 +73,13 @@
       },
 
       stopContainer: function () {
-        resources.Container.delete(this.stopContainerDialog.containerToStop)
+        let containerToStop = this.stopContainerDialog.containerToStop;
+        this.closeStopContainerDialog();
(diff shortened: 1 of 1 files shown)


frontend/admin/vue-components/ContainersView.vue
```diff
@@ -73,13 +73,13 @@
...
+        let containerToStop = this.stopContainerDialog.containerToStop;
+        this.closeStopContainerDialog();
+        resources.Container.delete(containerToStop)
         .done(() => {
           this.updateTable();
-          this.closeStopContainerDialog();
         })
         .fail(() => {
-          this.closeStopContainerDialog();
           this.communicationError = "The request could not be executed successfully";
```
stefanoborini: I don't recall where the communication error will go?

PR #528, opened 2017-06-22
Add Simphony logo to the login page

martinRenou:


stefanoborini:
We need a simphony logo and it must be smaller.

stefanoborini:
I would not put that file in jupyterhub. That directory is a "startup launchpad". Also, it won't be installed buy the setup that way.

martinRenou:
Should it be in the static directory of remoteappmanager ?

PR #529, opened 2017-06-22
Fix build by renaming apt docker dependency

martinRenou:
Apparently something has changed at the docker apt repository, so the result is that the environment setup is compromised.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ deps:
 		$(lsb_release -cs) \
 		stable"
 	-sudo apt-get -qq update
-	sudo apt-get install docker-ce=17.05.0~ce-0~ubuntu-$(lsb_release -cs)
+	sudo apt-get install docker-ce=17.03.0~ce-0~ubuntu-$(lsb_release -cs)
 	if [ `lsb_release -rs` = "14.04" ]; then \
 		plat_packages="linux-image-extra-$(uname -r) linux-image-extra-virtual python3.4-venv"; \
 	else \
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,11 @@ deps:
 		$(lsb_release -cs) \
 		stable"
 	-sudo apt-get -qq update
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -11,9 +11,9 @@ venv:
 .PHONY: dockerengine
 dockerengine:
```
stefanoborini: OK, at this point, I think we should 
- drop this rule, it's no longer relevant. You will also have to remove it from the docs (installation)
- Remove travis provided docker in the travis yaml file. 
- Have make deps do the docker installation (as it is doing now)

Makefile
```diff
@@ -25,13 +25,19 @@ deps:
 		false; \
 	fi
 	curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
+	curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+	sudo add-apt-repository \
+		"deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+		`lsb_release -cs` \
+		stable"
 	-sudo apt-get -qq update
 	if [ `lsb_release -rs` = "14.04" ]; then \
```
stefanoborini: ... don't bother with this one, because at this point the standard name is docker-engine for all versions.

martinRenou: I kept it because the installed version of python wasn't the same, it is python3.4-venv for trusty and python3-venv for other versions

martinRenou: Ok sorry I didn't understand this comment came with the other. I'll do it

Makefile
```diff
@@ -24,19 +14,17 @@ deps:
 		echo "ERROR: Cannot run on non-Linux systems"; \
 		false; \
 	fi
+	sudo apt-get remove docker docker-engine
```
stefanoborini: in the travis. We don't want to remove a docker installation on the user's machine. if there's already a docker, it will just upgrade it.

PR #530, opened 2017-06-26
Add column formatters for the datatable component

### LINKED ISSUE #476: Containers table in admin does not shorten the identifiers
The Containers table in the admin "Containers" tab has full length identifiers. This makes the table very long and hides the Stop button.

martinRenou:
Add `columFormatters` prop to the datatable component. This prop is an array of functions (formatters) which takes the value of the table as input and returns a string representing some html or plain text. This html output will be used to represent the data in the datatable.

Those formatters are used to shorten ids in the containersView and to represent boolean values in the accountingView.

Closes #476


martinRenou:
Maybe we should add a tooltip to display the complete identifier/name when the mouse is over it ?

stefanoborini:
simphonyproject should stay

martinRenou:
Need some tests

frontend/toolkit/DataTable.vue
```diff
@@ -14,8 +14,7 @@
         <tbody>
           <tr v-for="(row, row_index) in rows">
             <template v-for="(value, col_index) in row">
-              <td v-if="isBoolean(value)"><i class="fa fa-check" v-if="value"></i></td>
-              <td v-else>{{value}}</td>
+              <td v-html="columnFormatters[col_index] !== undefined ? columnFormatters[col_index](value) : value"></td>
```
martinRenou: check if it's a function.

frontend/toolkit/DataTable.vue
```diff
@@ -32,13 +31,14 @@
 
 <script>
   module.exports = {
-    props: [
-      "headers", "rows", "globalActions", "rowActions"
-    ],
+    props: {
+      headers: { type: Array, default: () => {return [];} },
```
martinRenou: must be mandatory

frontend/toolkit/DataTable.vue
```diff
@@ -32,13 +31,14 @@
...
+      columnFormatters: { type: Array, default: () => {return [];} },
+      rows: { type: Array, default: () => {return [];} },
```
martinRenou: i would be tempted to say it must also be mandatory, at least as model object from outside.

PR #552, opened 2021-07-13
MAINT: Update JupyterHub to v0.8.1

### LINKED ISSUE #397: Tests not passing against jupyterhub master
We currently pin against jupyterhub/jupyterhub@2d1a45f0190059ef436c2f97dc8d6e391eb2d139 . As of today (10 April) master brings some of our tests red.

### LINKED ISSUE #596: Don't rely on insecure cookies from JupyterHub to verify users
Currently the `remoteappmanager.services.hub.Hub ` class authenticates users and calls to the remoteappmanager REST API by extracting the access token from the `jupyter-hub-token-{user}` cookie and injecting it into the `Authorization` header of any outgoing requests.

https://github.com/simphony/simphony-remote/blob/15c24cb243a08e5b04c5444454be316ebf425dcc/remoteappmanager/services/hub.py#L36-L39

This relies on the JupyterHub cookies being visible on the client side by default (i.e. **NOT** HTTP secure), which introduces a security risk and also means that our authentication procedure is tightly coupled to JupyterHub's internals.

it's possible that this was simply a pragmatic development choice, as there may not have been any other way to handle authentication between the hub and the spawned application with early versions of JupyerHub (we are pinned at < 0.8.0). 

However, since JupyerHub moved to issuing HTTP only cookies in v0.8.0, we will not be able to support any later versions of the package until this is resolved.

flongford:
Closes #397
Closes #596 

This PR bumps the pinned version of `jupyterhub` to release 0.8.1

Changelog for 0.8.x series of JupyterHub: https://jupyterhub.readthedocs.io/en/stable/changelog.html#id44

Commit comparison between 0.8.1 and current pinned commit on `master`: https://github.com/jupyterhub/jupyterhub/compare/2d1a45f0190059ef436c2f97dc8d6e391eb2d139...0.8.1

Major changes include:

- Replacing `jupyterhub.orm.Proxy` with `jupyterhub.proxy.Proxy` where appropriate
- Replacing `jupyterhub.orm.Hub` with `jupyterhub.objects.Hub` where appropriate
- Updating handling of `BaseApplication` settings between `CommandLineConfig` and `EnvironmentConfig`
- Adding a `jupyterhub` versioning check to the `BaseApplication` class (may not be required)
- Authenticating with JupyterHub as an OAuth provider, rather than using (deprecated) session cookie


remoteappmanager/services/reverse_proxy.py
```diff
@@ -35,11 +33,11 @@ def __init__(self, *args, **kwargs):
...
 
-        # Note, we use jupyterhub orm Proxy, but not for database access,
-        # just for interface convenience.
-        self._reverse_proxy = jupyterhub_orm.Proxy(
+        # Note: we just use the jupyterhub ConfigurableHTTPProxy for
+        # interface convenience
+        self._reverse_proxy = ConfigurableHTTPProxy(
             auth_token=self.api_token,
-            api_server=_server_from_url(self.endpoint_url)
+            api_server=self.endpoint_url
```
flongford, line 40: Should be assignment of `api_url`

remoteappmanager/jupyterhub/spawners.py
```diff
@@ -32,24 +31,29 @@ def cmd(self):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-
         # We get the first one. Strangely enough, jupyterhub has a table
         # containing potentially multiple proxies, but it's enforced to
         # contain only one.
-        self.proxy = self.db.query(orm.Proxy).first()
+        self.proxy = self.db.query(ConfigurableHTTPProxy).first()
```
flongford, line 37: This won't work - we need to somehow get a reference to the application's `JupyterHub.proxy` attribute

flongford: Actually we only really need a reference to the Proxy API token and url

remoteappmanager/jupyterhub/spawners.py
```diff
@@ -25,31 +24,38 @@ class BaseSpawner(LocalProcessSpawner):
...
+        self.proxy = self.user.settings.get('proxy')
 
     def get_args(self):
         args = super().get_args()
 
-        for iarg, arg in enumerate(args):
-            args[iarg] = arg.replace('--base-url=', '--base-urlpath=')
+        # Handle no default IP value (removed in jupyterhub v0.8.0)
+        if not self.ip:
+            args.append('--ip="127.0.0.1"')
```
flongford, line 42: It's probably worth attempting to include a traitlet default here again, or if this fails document that there is a bug in `jupyterhub>0.8.0<0.9.0` that prevents doing so

remoteappmanager/base_application.py
```diff
@@ -161,22 +166,67 @@ def _registry_default(self):
...
+                yield gen.sleep(min(2 ** i, 16))
+            else:
+                break
+        else:
+            self.exit(1)
+
+        hub_version = resp.headers.get('X-JupyterHub-Version')
+        _check_version(hub_version, __version__, self.log)
+
+    def patch_default_headers(self):
```
flongford, line 219: Not: include docstring

jupyterhub/jupyterhub_config.py
```diff
@@ -9,16 +9,20 @@
 
 c.JupyterHub.hub_ip = public_ips()[0] if len(public_ips()) else '127.0.0.1'
 
+c.ConfigurableHTTPProxy.command = [
+    'configurable-http-proxy',
+    f'--default-target=http://{c.JupyterHub.hub_ip}:8081']
```
flongford, line 14: Note: this isn't really needed

remoteappmanager/application.py
```diff
@@ -22,5 +22,5 @@ def _web_handlers(self):
             )+"/?", RegisterContainerHandler),
             (base_urlpath, UserHomeHandler),
             (base_urlpath.rstrip('/'), web.RedirectHandler, {
-                "url": base_urlpath}),
+                "url": '/hub/user-redirect'}),
```
flongford, line 25: Note: this change doesn't do anything

flongford:
After playing around on the weekend, I've found the cause of the current redirect loop - the `HubAuthenticator.authenticate` method relies on the client-side `remoteappmanager` to be able to extract cookies issued by JupyterHub from incoming requests to verify the current user. This is not possible since JupyterHub switched to using HTTP only cookie protection in 0.8.0.

Therefore the user is not able to be authenticated with JupyterHub and gets caught in login loop.

### Discussion

First of all, we need to provide additional logging when a user is unable to be verified with JupyterHub and return a 401 / 403 if it happens, rather than redirecting.

Since the `BaseAuthenticator` already has user information we can also simply disable this check in `BaseHandler.prepare` and extract the user information from `self.application.user`. However, any requests to JupyterHub REST API (and therefore SR REST API) are still unverified, so we don't get very far.

Another option would be to verify the requests using JupyterHub's OAuth, as the `SingleUserNotebookApp` does. This set up is a bit more complicated than we need, since I believe Jupyter notebooks are their own OAuth providers, so we can't simply lift the code that performs the authentication step into our `BaseApplication`. However, we still might be able to use the same `authorization` and `token` OAuth endpoints.

Unfortunately we can't follow the same login procedure as a JupyterHub `Service`, since this expects the `jupyter-hub-services` cookie to be present, and since we are using a customised `Spawnner` instead, the hub issues a `jupyter-hub-token` cookie

### Proposal

- [ ] Provide additional logging when a user is unable to be verified with JupyterHub and return a 401 / 403 if it happens.
- [ ] Authenticate users with JupterHub using OAuth endpoints

nginx/nginx.conf.template
```diff
@@ -37,11 +37,7 @@ server {
 
 	# Managing literal requests to the JupyterHub front end
 	location / {
-	    # Note: this url uses https, since we are working on an older version of JupyterHub (< 0.8.0), which
-	    # internall uses the deprecated jupyterhub.orm.Proxy class. It is expected that updating the version
-	    # to > 0.8.0 will mean using a http address instead, since this is replaced by the
-	    # jupyterhub.proxy.ConfigurableHTTPProxy class instead
-		proxy_pass https://127.0.0.1:8000;
+		proxy_pass http://127.0.0.1:8000;
```
flongford, line 40: Note: need to test this on a remote deployment first

remoteappmanager/base_application.py
```diff
@@ -36,7 +40,7 @@ class BaseApplication(web.Application, LoggingMixin):
     #: API access to the configurable-http-proxy
     reverse_proxy = Instance(ReverseProxy)
 
-    #: API access to jupyterhub (for auth requests)
+    #: API access to JupyterHub for requests
```
flongford, line 43: Can undo this change

remoteappmanager/handlers/handler_authenticator.py
```diff
@@ -32,3 +35,23 @@ def authenticate(cls, handler):
...
+
+class HubOAuthenticator(Authenticator):
+    """Authenticator that uses the remote JupyterHub as an OAuth
+    provider to validate the request."""
+    @classmethod
+    @gen.coroutine
+    def authenticate(cls, handler):
+        # Authenticate the user against the hub. Expects the handler
+        # to inherit from the jupyterhub.services.auth.HubOAuthenticated
+        # mixin
```
flongford, line 48: Since we just use the `get_user` method, theoretically hander just needs to inherit from `HubAuthenticated`

PR #560, opened 2021-08-10
Extend Configurables to include a startup file

### LINKED ISSUE #555: Extend Simphony-Remote Configurables to specify an “input” file for application container
This should map to con environmental variable that is set in the container and can be dynamically assigned during runtime of a Simphony-Remote session.

It is expected that an “admin” user will assign the configurable to the image when registering the application. The value of the configurable should be exposed in the backend (maybe with a default value of 0), but does not need to be set in the frontend by the user (like the window size).

This needs to be well documented as its a sub-optimal approach - the image source code itself should really register the additional configurable object. Additionally, the "configurables" framework seems complex, so we may need some exploratory time here.

robertopreste:
Closes #555 

This PR adds a new Configurable, `ParaviewData`, that can exploit the `SRDATA` env variable exposed by Paraview-S-R to open a specific file when ParaView starts up. 

The Admin user can set a policy where this functionality is enabled, in which case the User will be shown a "Startup file" text input in the main view, where the path to a startup file can be provided. Currently, the only functional path is `/ParaView/can.ex2`, which is an example file included in the ParaView Docker image.


flongford (approved):
Couple of nits, otherwise LGTM!

robertopreste:
> Couple of nits, otherwise LGTM!

Right, thanks for catching these points, I totally overlooked them!

PR #566, opened 2021-09-03
Extend remoteapprest to add startupdata option

### LINKED ISSUE #565: Expand Simphony-Remote start command in rest API to include startupdata
The `remoteapprest app start` command should accept a `--startupdata` option, specifying the path to a file that would be used at startup, using the framework provided in #560. 
This probably involves editing the `tornado-webapi` module as well, since it would parse this option and send it back to Simphony-Remote to start the container.

robertopreste:
Closes #565 

The `remoteapprest app start` CLI command now accept a `--startupdata` option to provide a file path that will be loaded at startup. 

Note: the URL returned by the `remoteapprest app start` command cannot be used to open the container yet; there are two issues with it:

- the port referred in the URL is not correct
- the URL contains an extra `/api/v1/` that needs to be removed

The correct URL is the one printed in the logs; working on a way to return it through `remoteapprest`.

### DIFF
--- a/remoteappmanager/cli/remoteapprest/__main__.py
+++ b/remoteappmanager/cli/remoteapprest/__main__.py
@@ -152,8 +152,20 @@ def start(ctx, identifier, startupdata):
 
     payload_dict = dict(mapping_id=identifier)
     if startupdata is not None:
-        payload_dict.update(
-            configurables=dict(startupdata=dict(startupdata=startupdata)))
+        # First make sure that the allow_startup_data policy is True
+        request_url = urljoin(url,
+                              "/user/{}/api/v1/applications/".format(username))
+        response = requests.get(request_url, cookies=cookies, verify=False)
+        apps_data = json.loads(response.content.decode("utf-8"))
+        app_data = apps_data["items"][identifier]
+        allow_startup_data = app_data["image"]["policy"]["allow_startup_data"]
+        if allow_startup_data:
+            payload_dict.update(
+                configurables=dict(startupdata=dict(startupdata=startupdata)))
+        else:
+            raise click.ClickException(
+                "The 'allow_startup_data' policy is False for the current "
+                "user.\nExiting.")
 
     payload = json.dumps(payload_dict)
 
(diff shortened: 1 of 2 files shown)


remoteappmanager/cli/remoteapprest/__main__.py
```diff
@@ -137,19 +137,25 @@ def available(ctx):
 
 @app.command()
 @click.argument("identifier")
+@click.option("--startupdata", default=None)
 @click.pass_context
-def start(ctx, identifier):
+def start(ctx, identifier, startupdata):
     """Starts a container for application identified by IDENTIFIER.
+    If ``startupdata`` is provided, the container will open the given file at
+    startup.
```
flongford, line 145: Out of curiosity, what would happen in the situation that the application policy does not allow `startupdata` to be assigned but a file is provided in the CLI?

Presumably at the moment the `start` request should fail - I don't think there's much more we could do since we don't have access to the application database to check this beforehand.

remoteappmanager/cli/tests/test_remoteapprest.py
```diff
@@ -141,6 +141,18 @@ def test_app_start(self):
             self.assertEqual(mock_post.call_args[0][1],
                              json.dumps({"mapping_id": "1"}))
 
+            self._remoteapprest("app start 1 --startupdata=/test")
```
flongford, line 144: Is there anyway we can test the situation where the application policy for `1` does not allow `startupdata` to be assigned / is not supported?

flongford (commented):
Overall looks good, thanks @robertopreste - just a couple of comments

remoteappmanager/cli/remoteapprest/__main__.py
robertopreste: Good catch, strangely enough in that case the container starts anyway, with the provided startup data too! I'll see how we can prevent this.

robertopreste:
@flongford I fixed this, so that now the CLI would first check the `allow_startup_data` policy, and start the container with the given data only if that policy is True.

flongford (approved):
LGTM!

PR #568, opened 2021-09-06
Return usable URL from remoteapprest app start

### LINKED ISSUE #567: Return correct URL from remoteapprest app start
Follow-up from #565: 

The URL returned by the `remoteapprest app start` command cannot be used to open the container yet; there are two issues with it:

- the port referred in the URL is not correct
- the URL contains an extra `/api/v1/` that needs to be removed

robertopreste:
Closes #567 

[Includes changes from #566, changes will clean once that PR is merged]

Passing the right port to the response parsed by `remoteapprest app start` seems to be quite convoluted, so here is an approach which is a bit of a cheating, but it works and doesn't affect any other functionality. 

The container name is changed to include the right host port (e.g. "/simphony-remote_49165"), however this new name is not used internally (in fact it doesn't show in the logs); the container URL is still e.g. "http://127.0.0.1:49165/user/rpreste/containers/simphony-paraview/", because it is generated earlier in the process.

The new alternative URL is not functional, but it is sent in the response to the `remoteapprest app start` call, so it can then be parsed and rearranged to return the correct URL.

### DIFF
--- a/remoteappmanager/cli/remoteapprest/__main__.py
+++ b/remoteappmanager/cli/remoteapprest/__main__.py
@@ -4,7 +4,7 @@ import os
 import requests
 import requests.utils
 import json
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlsplit
 
 import click
 
@@ -161,6 +161,11 @@ def start(ctx, identifier, startupdata):
                              verify=False)
     if response.status_code == 201:
         location = response.headers["Location"]
+        parsed = urlsplit(location)
+        path, port = parsed.path.split("_")
+        port = port.rstrip("/")
+        path = "".join(path.split("/api/v1"))
+        location = f"{parsed.scheme}://{parsed.hostname}:{port}{path}"
         print(location)
 
 
--- a/remoteappmanager/cli/remoteapprest/__main__.py
+++ b/remoteappmanager/cli/remoteapprest/__main__.py
@@ -137,9 +137,12 @@ def available(ctx):
 
 @app.command()
 @click.argument("identifier")
+@click.option("--startupdata", default=None)
 @click.pass_context
-def start(ctx, identifier):
+def start(ctx, identifier, startupdata):
     """Starts a container for application identified by IDENTIFIER.
+    If ``startupdata`` is provided, the container will open the given file at
+    startup.
     If a container is already running, restarts it."""
     cred = ctx.obj.credentials
     url, username, cookies = cred.url, cred.username, cred.cookies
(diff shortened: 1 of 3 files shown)


remoteappmanager/cli/remoteapprest/__main__.py
```diff
@@ -137,24 +137,47 @@ def available(ctx):
...
     response = requests.post(request_url, payload, cookies=cookies,
                              verify=False)
     if response.status_code == 201:
         location = response.headers["Location"]
+        parsed = urlsplit(location)
+        path, port = parsed.path.split("_")
+        port = port.rstrip("/")
+        path = "".join(path.split("/api/v1"))
+        location = f"{parsed.scheme}://{parsed.hostname}:{port}{path}"
         print(location)
```
flongford, line 181: It would be useful to test the url parsing here to avoid regression. 

I guess we just need to check the stdout in the unittest runner, which I believe we can easily get from the `Result.output` object returned by `click.invoke`

remoteappmanager/docker/container_manager.py
```diff
@@ -349,7 +349,7 @@ def _start_container(self,
             '\n'.join('{0} -> {1}'.format(source, target['bind'])
                       for source, target in filtered_volumes.items()))
 
-        container_url_id = _generate_container_url_id()
+        container_url_id = image_name.split("/")[-1]
```
flongford, line 352: I am wondering if there are either any security or practical implications of this change, since the container url ID is no longer a random or unique string.

Are there any best practices for Docker container url ids? I couldn't find much information online. Maybe we can just review any potential information that we are exposing here

flongford (commented):
Thanks @robertopreste - I have a question about the potential implications of changing the container ID, but otherwise LGTM.

Just to clarify, is this PR currently is preferred over #567?

remoteappmanager/docker/container_manager.py
robertopreste: I agree, however I don't think we're exposing any sensible data, being it simply the name of the image. Also, it shouldn't pose a problem practically (e.g. name conflicts with other containers), since we run only one container (for a given image) at a time. 

I haven't found anything specific about container IDs, so I wouldn't know about best practices, I'm afraid.

robertopreste:
> Just to clarify, is this PR currently is preferred over #567?

This PR simply contains the changes from #567, because the new changes rely on them. I opened two separate PRs only to keep things in order; even if it's very small changes I think it would be better to review and merge them separately.

remoteappmanager/docker/container_manager.py
flongford: Yes, that makes sense.

To tidy things up, could we extend the `_generate_container_url_id` to take a `image_name` keyword argument that is `None` by default? If the `image_name` is then provided, then we parse it, if not we just return a random string.

flongford (approved):
Thanks for your responses @robertopreste - LGTM!

PR #569, opened 2021-09-06
Fix User demo application assignment

flongford:
Small fixes and enhancements required to allow a set of demo applications provided to all new users to be declared in the `jupyterhub/remoteappmanager_config.py` .

This is achieved by introducing two features:
- `FileConfig.demo_applications` can be configured with a list of image names that are already registered with the application
- `FileConfig.auto_user_creation` can be configured to allow automatic registration of any authenticated user. If set to `True`, as long as the user can login via the declared `c.JupyterHub.authenticator_class`, then an account will be automatically created in the Simphony-Remote database. Otherwise an admin user will have to manually perform the task

Examples of how these may be assigned are given in `jupyterhub/remoteappmanager_config.py`

### DIFF
--- a/remoteappmanager/base_application.py
+++ b/remoteappmanager/base_application.py
@@ -168,17 +168,18 @@ class BaseApplication(web.Application, LoggingMixin):
     # Private
     def _add_demo_apps(self, user):
         """Grant access to any demo applications provided for user"""
-
-        if user.demo_applications:
+        if not user.demo_applications:
+            self.log.debug("No demo applications available")
             return
 
         # Add all demo applications already registered
         for application in self.db.list_applications():
             if application.image in user.demo_applications:
-                self.log.info(application.image)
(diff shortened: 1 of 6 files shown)


remoteappmanager/base_application.py
```diff
@@ -168,20 +180,26 @@ def start(self):
...
+        if not self.file_config.demo_applications:
+            self.log.debug("No demo applications available")
             return
 
         # Add all demo applications already registered
         for application in self.db.list_applications():
-            if application.image in user.demo_applications:
-                self.log.info(application.image)
+            if application.image in self.file_config.demo_applications:
+                self.log.debug(f"Avaliable image: {application.image}")
```
robertopreste, line 194: Just a small typo here.

robertopreste (approved):
Looks very good, thanks for the changes!

PR #612, opened 2022-06-29
Expose demo application policy options in config

### LINKED ISSUE #611: Expose demo applications policy options
Developers have the ability to provide a list of "demo" or "default" application names that are available to all users upon login via the `demo_applications` entry in the `remoteappmanager_config.py` configuration file.

However, it would be advantageous to also be able to configure the policy options that are set for these applications, since currently these are not exposed.

I think the most elegant way might be to convert the `demo_applications` list into a dictionary (keyed by app name), allowing users to provide overrides to the default options for each case.

flongford:
Closes #611

Follows suggestion in #611 and exposes demo application policy options as a dictionary in `remoteappmanager_config.py`

### DIFF
--- a/remoteappmanager/base_application.py
+++ b/remoteappmanager/base_application.py
@@ -21,7 +21,7 @@ from remoteappmanager.services.reverse_proxy import ReverseProxy
 
 
 DEFAULT_POLICY_OPTIONS = {
-    "app_license": '',
+    "app_license": None,
     "allow_home": False,
     "allow_view": True,
     "volume": None,
--- a/remoteappmanager/base_application.py
+++ b/remoteappmanager/base_application.py
@@ -20,6 +20,15 @@ from remoteappmanager.services.hub import Hub
 from remoteappmanager.services.reverse_proxy import ReverseProxy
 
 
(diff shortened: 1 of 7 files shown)


remoteappmanager/base_application.py
```diff
@@ -20,6 +20,15 @@
 from remoteappmanager.services.reverse_proxy import ReverseProxy
 
 
+DEFAULT_POLICY_OPTIONS = {
+    "app_license": '',
```
flongford:
```suggestion
    "app_license": None,
```

remoteappmanager/tests/test_application.py
```diff
@@ -109,7 +110,9 @@ def test_initialization_with_demo_applications(self):
             "remoteappmanager.db.orm.ORMDatabase")
         self.file_config.database_kwargs = {
             "url": "sqlite:///"+sqlite_file_path}
-        self.file_config.demo_applications = ['my-demo-app']
+        self.file_config.demo_applications = {
+            'my-demo-app': {}
```
flongford: Would be beneficial to test an override here, maybe of `allow_home` option

